PR #19, opened 2018-06-11
Remove hardcoded path to Android SDK directory

MichaelEvans:


danielbuechele:
Can we have a fallback to `opt/android_sdk` like this:
```js
'${ANDROID_HOME:=/opt/android_sdk}/tools/emulator @' + name
```
For some reason on my machine ANDOIRD_HOME isn't set and otherwise the command fails. I think this is due to some Facebook specific setup.

xiphirx:
That sounds like a bad config issue on FB's side. `ANDROID_HOME` is the standard env var to point to the SDK directory, and all the standard Android tools use it. If we want a fallback, it should instead fallback to `ANDROID_SDK_ROOT`.

Issue #22, opened 2018-06-11
Fix `Getting Started` docs Android 


andreiverdes:
`Getting Started` Android docs is using an internal private method to check if Sonar should be used.
As per GH-14 comment, it should point to `SonarUtils.shouldEnableClient(this)`

Issue #14, opened 2018-06-11
error: isMainProcess(Context) has private access in SonarUtils


andreiverdes:
Pretty explanatory from the title :D 
Trying to setup Sonar and I'm having a compile issue

Rohfosho:
You can use `SonarUtils.shouldEnableClient(this)` instead. 

This checks if it's the main process, an end to end test, and if it's an internal build.

andreiverdes:
Oh okay, then :) the getting started guide was using that, closing this issue

PR #23, opened 2018-06-11
Fix GH-22 wrong method for Android docs

andreiverdes:
As per GH-14 comment, the docs were pointing to an internal `private` method to check if `Sonar` should be enabled

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -65,7 +65,7 @@ public class MyApplication extends Application {
   public void onCreate() {
     super.onCreate();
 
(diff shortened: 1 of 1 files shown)


rigor789:
I just checked and I believe that should be `SonarUtils.shouldEnableSonar(this)`

https://github.com/facebook/Sonar/blob/1cc2d46be9097eb73cf99f300105c1a719b908d9/android/android/utils/SonarUtils.java#L19-L21

andreiverdes:
My bad!

PR #52, opened 2018-06-12
Add missing Android dependencies

MichaelEvans:
Seems like the Android build depends on Rhino but the dependency is not declared, and there's a `/res` directory listed in the sourceSets, but the directory doesn't exist. I _think_ this should cover those cases (and delete a duplicate guava dependency in the deps list).


priteshrnandgaonkar:
The work related to fixing android dependencies is being done in Fix-Android branch. Some of this changes are incorporated in that branch. So this PR is not required for now. But thanks for giving a try.

PR #51, opened 2018-06-12
Update .gitignore to better support Android / IntelliJ config

MichaelEvans:

PR #55, opened 2018-06-12
Remove local.properties and ignore it

xiphirx:
This file is specific to the machine the project runs on and should not be checked in.


MichaelEvans:
@xiphirx Added in #51 :)

Issue #53, opened 2018-06-12
Docs translation


romainwn:
Is there any plans for internationalisation ?
I would love to help translate the docs

danielbuechele:
We can not provide this internally. Community translations are definitely appreciated. We are using Docusaurus to generate our website, which supports [translations](https://docusaurus.io/docs/en/translation).
One thing I want to make sure when adding translations is that the translations are well maintained and correct. It could become super confusing, when we changing APIs and change the English docs, but other languages still refer to older versions.

Which language would you be able to provide?

romainwn:
I definitely take a look at Docusaurus in the next few days :) 

I know that Electron uses [Crowdin](https://crowdin.com/project/electron) but I didn't get a chance to take a good look at it yet.

I would be able to provide the French part !

passy:
Closing this as we've decided against hosting/supporting multiple languages for now.

PR #68, opened 2018-06-13
[WIP] setting up docs translation

romainwn:
I looked into [Docusaurus docs](https://docusaurus.io/docs/en/translation) and started a PR related to #53 
I actually don't know if this is still WIP or almost ready to be merged, but I am pretty sure that we have to link the project to Crowdin ([doc here](https://docusaurus.io/docs/en/translation#crowdin))

Please let me know what you think of this PR :)


passy:
Apologies for the long wait and thanks for the PR, @romainwn. We have given this some thought and come to the conclusion that, at this point, we are not ready to support multiple languages. This is something we may want to revisit at a later point in time for this project, but right now we want to focus on different areas first. Thanks again!

PR #91, opened 2018-06-19
Fix parameters bug

x4snowman:
facebook::cxxutils::convertIdToFollyDynamic can only accept one paramter


noahsark769:
++, also seeing this bug when pulling in SonarKit

priteshrnandgaonkar:
We are working on fixing the sync of FBCxxUtils. The fix is coming soon.

jknoxville:
Hey, thanks for raising this, and thanks for the pull request though there is a reason that we need that boolean parameter. There was an issue with code syncing where the dependency didn't get synced to github but the usage of it did.

I've fixed it in 94f03f182b44bb658217ee38071702e375436883

PR #97, opened 2018-06-20
Fix Sample iOS App to compile and execute

pkrmf:
The iOS Sonar Sample App is unable to compile due to not having the ComponentKit podspec updated.

- [X] ComponentKit was declaring a dependency on Yoga 1.6.0. We should never declare a dependency to a patch version, only minor. Cocoapods would treat 1.6.0 to 1.7.0 as a major version change, when it is not.

- [X] ComponentKit.podspec was defining version 0.21 which didn't include enum values that SonarKit depends on, causing the sample app to fail to compile. ComponentKit 0.22 tag is also missing some commits. As of now, the local ComponentKit.podspec won't be pulling 0.22 tag, instead, it will pull master, which allows the app to compile and execute.

### DIFF
--- a/iOS/third-party-podspecs/ComponentKit.podspec
+++ b/iOS/third-party-podspecs/ComponentKit.podspec
@@ -1,12 +1,12 @@
 Pod::Spec.new do |s|
   s.name = 'ComponentKit'
-  s.version = '0.21'
+  s.version = '0.22'
   s.license = 'BSD'
   s.summary = 'A React-inspired view framework for iOS'
   s.homepage = 'https://componentkit.org'
   s.social_media_url = 'https://twitter.com/componentkit'
   s.authors = 'adamjernst@fb.com'
-  s.source = { :git => 'https://github.com/facebook/ComponentKit.git'}
+  s.source = { :git => 'https://github.com/facebook/ComponentKit.git' }
   s.ios.deployment_target = '8.1'
   s.requires_arc = true
 
@@ -17,5 +17,5 @@ Pod::Spec.new do |s|
     'CLANG_CXX_LANGUAGE_STANDARD' => 'gnu++14',
     'CLANG_CXX_LIBRARY' => 'libc++',
   }
-  s.dependency 'Yoga'
+  s.dependency 'Yoga', '~> 1.6'
 end
--- a/iOS/Sample/Podfile.lock
+++ b/iOS/Sample/Podfile.lock
@@ -1,53 +1,62 @@
 PODS:
   - boost-for-react-native (1.63.0)
   - CocoaAsyncSocket (7.6.3)
-  - ComponentKit (0.21):
-    - Yoga
+  - CocoaLibEvent (1.0.0)
+  - ComponentKit (0.22):
+    - Yoga (~> 1.6)
   - DoubleConversion (3.0.0)
-  - EasyWSClient (1.0.0)
-  - Folly (2018.05.07.00):
+  - Folly (1.0.0):
     - boost-for-react-native
+    - CocoaLibEvent (~> 1.0)
     - DoubleConversion
     - glog
+    - OpenSSL-Static (= 1.0.2.c1)
   - glog (0.3.5)
-  - PeerTalk (0.0.2)
-  - Sonar (1.0.0):
-    - EasyWSClient
+  - OpenSSL-Static (1.0.2.c1)
+  - PeerTalk (1.0.0)
+  - RSocket (0.10.0):
     - Folly
-  - SonarKit (1.0.0):
+  - Sonar (0.0.1):
+    - Folly
+    - RSocket
+  - SonarKit (0.0.1):
     - CocoaAsyncSocket (~> 7.6)
     - Folly
+    - OpenSSL-Static (= 1.0.2.c1)
     - PeerTalk
     - Sonar
-    - SonarKit/SKIOSNetworkPlugin (= 1.0.0)
-    - SonarKit/SonarKitLayoutComponentKitSupport (= 1.0.0)
-    - SonarKit/SonarKitLayoutPlugin (= 1.0.0)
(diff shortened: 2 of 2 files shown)


iOS/third-party-podspecs/ComponentKit.podspec
```diff
@@ -17,5 +17,5 @@ Pod::Spec.new do |s|
     'CLANG_CXX_LANGUAGE_STANDARD' => 'gnu++14',
     'CLANG_CXX_LIBRARY' => 'libc++',
   }
-  s.dependency 'Yoga'
+  s.dependency 'Yoga', '~> 1.6'
```
priteshrnandgaonkar: This should be `1.8`. https://cocoapods.org/pods/Yoga

iOS/third-party-podspecs/ComponentKit.podspec
```diff
@@ -1,12 +1,12 @@
...
   s.name = 'ComponentKit'
-  s.version = '0.21'
+  s.version = '0.22'
   s.license = 'BSD'
   s.summary = 'A React-inspired view framework for iOS'
   s.homepage = 'https://componentkit.org'
   s.social_media_url = 'https://twitter.com/componentkit'
   s.authors = 'adamjernst@fb.com'
-  s.source = { :git => 'https://github.com/facebook/ComponentKit.git'}
+  s.source = { :git => 'https://github.com/facebook/ComponentKit.git' }
```
priteshrnandgaonkar: nit: Extra space

iOS/third-party-podspecs/ComponentKit.podspec
```diff
@@ -1,12 +1,12 @@
 Pod::Spec.new do |s|
   s.name = 'ComponentKit'
-  s.version = '0.21'
+  s.version = '0.22'
```
priteshrnandgaonkar: I don't understand the change u did in `ComponentKit.podspec`. AFAIK if u don't mention anything like `:branch`, `:tag` in `s.source` it will pull in from master. So changing the version number from `0.21` to `0.22` will not serve any purpose. It will still pull from master.

pkrmf: ComponentKit 0.22 tag is missing some commits that include enum values SonarKit depends on. The only way to get sonarkit to compile is by depending on ComponentKit latest commit(master). We could just define the dependency on the latest commit or open an issue in ComponentKit to update its 0.22 tag or potentially create a new tag.

iOS/third-party-podspecs/ComponentKit.podspec
pkrmf: I added 1.6 since ComponentKit podspec in componentkit repo points to 1.6.0. I will change it to 1.8 since it should have no impact

iOS/third-party-podspecs/ComponentKit.podspec
priteshrnandgaonkar: I did `pod install` with the current setup and for componentkit it pulled the code from current master. So all the required commits will be available. By setting `s.version = '0.21'` doesn't check out `0.21` tag since its not mentioned in the `s.source` argument. To checkout `0.21` tag, you will have to add `:tag` argument. Checkout [glog podspec](https://github.com/facebook/Sonar/blob/master/iOS/third-party-podspecs/glog.podspec). So tecchnically with current setup you can assign any value `s.version` still it will checkout master.

pkrmf: Sorry I didn’t quite understand you. When you say current setup, you mean my pull request or master?

I am aware that with my setup, if you don’t define a tag in the s.source it will pull master. Setting a tag, either 0.21 or 0.22 will cause the sample app to not build.

Maybe I misunderstood you.

priteshrnandgaonkar: > Sorry I didn’t quite understand you. When you say current setup, you mean my pull request or master?
By current setup I mean master.

> I am aware that with my setup, if you don’t define a tag in the s.source it will pull master. Setting a tag, either 0.21 or 0.22 will cause the sample app to not build.

True. So in order to not depend on master, componentkit guys have to release the new pod in Cocoapod. So in your current PR doing `s.version = '0.22'` still pulls master. So this change is not required, right?

pkrmf: I think it would be good to get this out there in the meantime. I opened an issue in componentkit repo so they can address it. In the meantime we could get this merge so people playing with sonarkit and its sample app are able to build and run it.

As of today we are pulling componentkit from a local podspec in this repo. Ideally we should be pointing to a version in the podfile and rely on cocoapods master to do its job.

priteshrnandgaonkar: >  In the meantime we could get this merge so people playing with sonarkit and its sample app are able to build and run it.

I can build sample app on master

pkrmf: Oh wow. You can build master without any changes? I am unable to. I will double check when I get home.

Did you pod install? Pod update?

priteshrnandgaonkar: I did pod install

pkrmf: @priteshrnandgaonkar I just pulled master, pod repo update, pod install and I get:
`marcterns@Marcs-MBP ~/iOS/Sonar/iOS/Sample (master) $ pod install
Analyzing dependencies
Pre-downloading: `ComponentKit` from `https://github.com/facebook/ComponentKit.git`, commit `89c5e28e4ccc02927187a8aac6b7196f533cf86d`
Pre-downloading: `Folly` from `https://github.com/facebook/folly.git`, commit `1ed5d00ca4f43d5c81f4d731119e71509179dc99`
[!] Unable to find a specification for 'Folly'.`

PR #98, opened 2018-06-22
Update the ios platform form 8.0 to 8.4, as the PeerTalk has a minimu…

x4snowman:
The PeerTalk in the Sonar project has a minimum platform requirement as 8.4


x4snowman:
Close, All dependency issues are fixed.

PR #104, opened 2018-06-26
[gradle] Set up Maven packaging and Bintray upload

passy:
Summary:
This make use of the the setup we have at Litho, which should make
it easier to keep them in sync.

It creates proper source jars, AARs (without bundling in unwanted
transitive deps) and POMs.

Test Plan:
```
./gradlew :android:bintrayUpload -info -PdryRun=true
```

Uploaded an `0.0.2` to bintray and it looks pretty good: https://jcenter.bintray.com/com/facebook/sonar/sonar/0.0.2/

NB: I'm not sure whether or not to update the license headers here. We could set up a sync between Litho and this which would break that.


xiphirx:
Do we really need the bintray stuff? The gradle-mvn-push script should suffice for everything no?

passy:
We publish to bintray, though. :)

Issue #93, opened 2018-06-20
Refactor Sonar Podspec to include all of its dependencies


pkrmf:
We should make sure to declare all the Sonar dependencies within Sonar's podspec, and make sure they point to the versions they depend on(we could also use the ~> for floating versions). 

It is very hard to maintain an application consuming Sonar since its Podfile has to manually define all Sonar's dependencies and the paths to the each one of the podspecs. What happens if we update Sonar? Then we also might have to manually update the dependencies Sonar is depending on.

There are two ways to prevent this from happening:
- Upload all the Sonar's dependencies podspecs to the cocoapods public master repo along with Sonar podspec and make sure we define those dependencies in the Sonar's podspec.
- Create a private repo and push the Sonar's podspec and its podspec dependencies to it. Then communicate to consumers of Sonar that they have to include the source of the private repo within the Podfile.

Sonar's consumers would go from this:

```
platform :ios, '8.0'
swift_version = '4.1'

target 'MyApp' do

  pod 'RSocket', :podspec => 'https://raw.githubusercontent.com/facebook/Sonar/master/iOS/third-party-podspecs/RSocket.podspec'
  pod 'DoubleConversion', :podspec => 'https://raw.githubusercontent.com/facebook/Sonar/master/iOS/third-party-podspecs/DoubleConversion.podspec'
  pod 'glog', :podspec => 'https://raw.githubusercontent.com/facebook/Sonar/master/iOS/third-party-podspecs/glog.podspec'
  pod 'Folly', :podspec => 'https://raw.githubusercontent.com/facebook/Sonar/master/iOS/third-party-podspecs/Folly.podspec'
  pod 'PeerTalk', :podspec => 'https://raw.githubusercontent.com/facebook/Sonar/master/iOS/third-party-podspecs/PeerTalk.podspec'
  pod 'Yoga','~>1.8.1', :modular_headers => true
  pod 'Sonar', :podspec => 'https://raw.githubusercontent.com/facebook/Sonar/master/xplat/Sonar/Sonar.podspec'
  pod 'SonarKit', :podspec => 'https://raw.githubusercontent.com/facebook/Sonar/master/iOS/SonarKit.podspec'
  pod 'SonarKit/SonarKitLayoutComponentKitSupport', :podspec => 'https://raw.githubusercontent.com/facebook/Sonar/master/iOS/SonarKit.podspec'
[...]

PR #107, opened 2018-06-26
CocoaPods Private Repo support

pkrmf:
This PR brings us one step closer for publishing SonarKit.podspec to the public Cocoapods master repository. This solves [#93](https://github.com/facebook/Sonar/issues/93).

- [X] `SonarKit.podspec` now lints by passing the `--sources` flag and the `--use-libraries` flag. `pod spec lint  SonarKit.podspec --sources=https://github.com/facebook/Sonar,master --allow-warnings --use-libraries` Same for `pod repo push`.

- [X] `SonarKit.podspec` is now also published to the CocoaPods Private repo. What does this mean? It means that we no longer need to manually define all of `SonarKit` dependencies in the Podfile. It only takes `SonarKit` consumers to add this line `source 'https://github.com/facebook/Sonar.git'` on top of any Podfile, and `SonarKit` will be installed by just defining it with `pod SonarKit`. 

- [X] We are publishing a Cocoapods Private Repo in the meantime we don't have all of our dependencies updated and published to the CocoaPods master repo. The CocoaPods Private Repo contains updated dependency podspecs that will need to be published to the CocoaPods master repo in the near future. That will be the next action item in order to have SonarKit.podspec published as well.

- [X] Sample App Podfile has been refactored, much simpler and cleaner.

- [X] SonarKit Framework project now pulls its dependencies from the cocoapods master repository and the temporary private repository instead of the local podspecs. I am able to compile the SonarKit framework project now, before I was unable to.
 
- [X] Local third party podspec dependencies have been removed.  

@emilsjolander @priteshrnandgaonkar @noahsark769 feel free to contribute or expose any concerns.


Specs/README.md
```diff
@@ -0,0 +1,2 @@
+# cocoapods-private-repo
+A private repo to develop Sonar
```
noahsark769: It might be good to include a little documentation here about how to publish new versions of specs on this private repo. Are we copying the files into the `Specs` directory to "publish" them, or using `pod repo push`?

pkrmf:
@priteshrnandgaonkar I will be removing the .DS_Store files this afternoon.

Also, good news! My PR for ComponentKit.podspec was successfully published to the cocoapods master repo, so I will be removing that from our private repo and pointing directly to master

pkrmf:
@priteshrnandgaonkar I made the required changes. I also added this:

- [X] SonarKit Framework project now pulls its dependencies from the cocoapods master repository and the temporary private repository instead of the local podspecs. I am able to compile the SonarKit framework project now, before I was unable to.

pkrmf:
@priteshrnandgaonkar Anything else I can help with in order to get this PR merged?

priteshrnandgaonkar:
@pkrmf, Nothing from your side, PR looks good. Was little busy. Importing it into internal system and will see how our internal tests respond.

priteshrnandgaonkar:
@pkrmf Any particular reasons for custom Yoga and YogaKit pods ?

noahsark769:
btw, I think we might be able to get around the `DEFINES_MODULE` issue by using `modular_headers => true` for the `Yoga` pod definition in `SonarKit.podspec` (although making Yoga's podspec define a module explicitly is probably a better solution)

pkrmf:
>btw, I think we might be able to get around the `DEFINES_MODULE` issue by using `modular_headers => true` for the `Yoga` pod definition in `SonarKit.podspec` (although making Yoga's podspec define a module explicitly is probably a better solution)

@noahsark769 That is also an option. I always try to make the adoption as easier as possible for consumers. If I can fix it in the podspec itself and avoid consumers to make more modifications to their podfiles, the better.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #114, opened 2018-06-28
Make Java package structure more reasonable

kele86838437:


passy:
Thanks for the suggestion, but this is based on internal structure. Also, for future contributions, please provide a description of what you're trying to achieve.

Issue #24, opened 2018-06-11
Gradle error when building


rigor789:
I'm trying to test Sonar with android, I created a new application with android studio, and added the dependency to the `build.gradle`

```gradle
dependencies {
    // ...
    debugImplementation 'com.facebook.sonar:sonar:0.0.1'
}
```

When I try to build the app, I get the following error:
```
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':app:transformDexArchiveWithExternalLibsDexMergerForDebug'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:100)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:70)
	at org.gradle.api.internal.tasks.execution.OutputDirectoryCreatingTaskExecuter.execute(OutputDirectoryCreatingTaskExecuter.java:51)
	at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:62)
	at org.gradle.api.internal.tasks.execution.ResolveTaskOutputCachingStateExecuter.execute(ResolveTaskOutputCachingStateExecuter.java:54)
	at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:60)
	at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:97)
	at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:87)
	at org.gradle.api.internal.tasks.execution.ResolveTaskArtifactStateTaskExecuter.execute(ResolveTaskArtifactStateTaskExecuter.java:52)
	at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:52)
	at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:54)
	at org.gradle.api.internal.tasks.execution.ExecuteAtMostOnceTaskExecuter.execute(ExecuteAtMostOnceTaskExecuter.java:43)
[...]

PR #115, opened 2018-06-28
[android] Exclude `libsonarfb.so` from sonar AAR

passy:
The default, even for libraries, is to include all downstream dependencies in the created AAR. This causes us to effectively provide `libsonarfb.so` twice, which creates unnecessary pain (see #24).

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -235,6 +235,10 @@ android {
         }
     }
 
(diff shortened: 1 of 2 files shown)


passy:
I think I need to reverse this. Because we link against specific symbols this doesn't seem to work.

PR #129, opened 2018-07-03
[android] v0.10.0

### LINKED ISSUE #111: Support x86_64 architecture
Currently we doesn't support x86_64 architecture for android, as we don't have prebuilt `so` file for openssl for that architecture. The best way to support it would be to get rid off all the prebuilt so files and use gradle to build it and link with sonar native targets.

One can follow https://wiki.openssl.org/index.php/Android#OpenSSL_Library

passy:
Summary:
New release including x86_64 binaries.

Depends on #128.
Closes #111.

Test Plan:
Sample app.


android/third-party/Folly/CMakeLists.txt
```diff
@@ -47,6 +47,11 @@ add_compile_options(
     -Wno-return-type
     -Wno-tautological-constant-compare
     )
+
+if (CMAKE_ANDROID_ARCH_ABI STREQUAL "x86_64")
```
priteshrnandgaonkar: Great work on this

PR #131, opened 2018-07-04
Keep strong reference to OnSharedPreferenceChangeListener

ronocod:
A strong reference to the listener should be kept when calling `registerOnSharedPreferenceChangeListener`, otherwise the listener can get garbage-collected and stop working.

Mentioned in the Android SDK docs [here](https://developer.android.com/reference/android/content/SharedPreferences.html#registerOnSharedPreferenceChangeListener(android.content.SharedPreferences.OnSharedPreferenceChangeListener)):

> Caution: The preference manager does not currently store a strong reference to the listener. You must store a strong reference to the listener, or it will be susceptible to garbage collection. We recommend you keep a reference to the listener in the instance data of an object that will exist as long as you need the listener.

### DIFF
--- a/android/plugins/sharedpreferences/SharedPreferencesSonarPlugin.java
+++ b/android/plugins/sharedpreferences/SharedPreferencesSonarPlugin.java
@@ -23,6 +23,21 @@ public class SharedPreferencesSonarPlugin implements SonarPlugin {
 
   private SonarConnection mConnection;
   private final SharedPreferences mSharedPreferences;
+  private final SharedPreferences.OnSharedPreferenceChangeListener onSharedPreferenceChangeListener = new SharedPreferences.OnSharedPreferenceChangeListener() {
+    @Override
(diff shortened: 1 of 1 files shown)


android/plugins/sharedpreferences/SharedPreferencesSonarPlugin.java
```diff
@@ -23,6 +23,21 @@
 
   private SonarConnection mConnection;
   private final SharedPreferences mSharedPreferences;
+  private final SharedPreferences.OnSharedPreferenceChangeListener onSharedPreferenceChangeListener = new SharedPreferences.OnSharedPreferenceChangeListener() {
+    @Override
+    public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
+      if (mConnection != null) {
```
xiphirx: I know that this was copied over, but can we early return here instead of indenting the entire block?

```
if (mConnection == null) {
  return;
}
```

PR #136, opened 2018-07-04
Android travis

priteshrnandgaonkar:


.travis.yml
```diff
@@ -60,3 +60,31 @@ matrix:
         - cd iOS/Sample
         - xcodebuild -showsdks
         - xcodebuild clean build -workspace Sample.xcworkspace -scheme Pods-Sample -sdk iphonesimulator11.4
+
+    - language: android
```
passy: Awesome! As a next step (not necessarily part of this), could we run this on Linux hosts to speed things up. There's actually support for [multi OS testing](https://docs.travis-ci.com/user/multi-os/) that I didn't know about and we could simply skip Android builds on MacOS and vice versa.

.travis.yml
```diff
@@ -60,3 +60,31 @@ matrix:
...
+
+            # Additional components
+            - extra-google-google_play_services
+            - extra-google-m2repository
+            - extra-android-m2repository
+            - addon-google_apis-google-27
+
+            # Specify at least one system image,
+            # if you need to run emulator(s) during your tests
+            - sys-img-armeabi-v7a-android-27
```
passy: I don't think we'll need this.

priteshrnandgaonkar: Yeah, but travis is not able to parse it https://travis-ci.org/facebook/Sonar/requests

.travis.yml
```diff
@@ -60,3 +60,33 @@ matrix:
         - cd iOS/Sample
         - xcodebuild -showsdks
         - xcodebuild clean build -workspace Sample.xcworkspace -scheme Pods-Sample -sdk iphonesimulator11.4
+
+    - language: android
+        jdk:
```
passy: This isn't consistently indented causing the parsing error.

.travis.yml
priteshrnandgaonkar: I run android builds on linux.

.travis.yml
```diff
@@ -60,3 +60,38 @@ matrix:
...
+        licenses:
+          - android-sdk-preview-license-.+
+          - android-sdk-license-.+
+          - google-gdk-license-.+
+      before_install:
+        - yes | sdkmanager "platforms;android-27"
+        - echo "y" | sdkmanager "cmake;3.6.4111459"
+        - wget https://dl.google.com/android/repository/android-ndk-r17b-linux-x86_64.zip
+        - mkdir -p temp-android-ndk
+        - mv ./android-ndk-r17b-linux-x86_64.zip temp-android-ndk
```
passy: Did installing the `ndk-bundle` via the sdk manager not work?

.travis.yml
```diff
@@ -60,3 +60,38 @@ matrix:
...
+          - extra-google-m2repository
+          - extra-android-m2repository
+          - addon-google_apis-google-27
+
+        licenses:
+          - android-sdk-preview-license-.+
+          - android-sdk-license-.+
+          - google-gdk-license-.+
+      before_install:
+        - yes | sdkmanager "platforms;android-27"
```
passy: Do we need the `android-27` target from above still?

Issue #79, opened 2018-06-15
Network plugin + OkHttp interceptor does not decompressed gzipped responses


damianw:
Responses with `Content-Type: gzip` are not unzipped, and render garbage as plaintext in the Sonar network view.

danielbuechele:
There is some logic in place to unzip requests, however for some reasons it is failing: https://github.com/facebook/Sonar/blob/master/src/plugins/network/RequestDetails.js#L64

@damianw Can you see why this is happening?

damianw:
I'm not exactly sure as I haven't had time to investigate it. In the meantime, I've added a custom interceptor to perform the unzipping on the Android side.

amenon:
@damianw typically, gzipped responses have the `Content-Encoding` set to `gzip` rather than the `Content-Type`.

I noticed that the plugin was not decompressing my OkHttp interceptor responses because my responses used a lower-case header name `content-encoding`. I have a PR (#143) to address that.

damianw:
My mistake, I meant `Content-Encoding`

passy:
I believe that this was fixed quite a while ago in the referenced commit.

PR #143, opened 2018-07-07
Ignore header case to determine if content is gzipped

amenon:
This PR updates the Network plugin's response parser to ignore the case of the `Content-Encoding` header in the response. This was preventing responses that are gzipped from being uncompressed. I tested this with gzipped responses through the OkHttp interceptor on Android. This could potentially address #79 as I was experiencing that issue before this change.

sonar


amenon:
@passy that make sense. I didn't see any other places in the code where headers were being examined. I assume that we would still want to display the raw headers in the UI and not the normalized version.

Feel free to convert this into an issue. Thanks!

PR #192, opened 2018-07-26
Fix Obj-C and Swift builds which are broken in v0.6.13

### LINKED ISSUE #190: Builds for non-Objc++ iOS apps are broken on v0.6.13
See [this repo](https://github.com/noahsark769/sonar-testing-example-projects) for two examples of apps which integrate Sonar in the recommended way but don't build. The error looks like this:

image

I believe this is because 9f807cd0 introduced the `CppBridge` subspec, and as a result C++ headers are getting included in SonarKit-umbrella.h. The Sample project still compiles because `AppDelegate.mm` is Obj-C++.

I've got a fix for this coming out soon :)

noahsark769:
9f807c introduced a new subspec in SonarKit.podspec, which caused Cpp header files to be included in `SonarKit-umbrella.h`, causing all non-ObjC++ project integrations to fail to build (see #190 for more info). @SirArkimedes and I investigated this today and found the fix:

1. Make the `CppBridge` subspec's header files private
2. Add a hack to make Cocoapods copy `FBMacros.h` into `{PODS_ROOT}/Headers/Public/SonarKit`, since otherwise `#import "FBMacros.h"` in `SonarKit-umbrella.h` will fail to resolve in out-of-the-box projects.

I've put up a branch [here](https://github.com/noahsark769/sonar-testing-example-projects/tree/working_reference) which pulls the pod from this branch and verifies that the fix works.

Fixes #190


iOS/SonarKit.podspec
```diff
@@ -55,7 +57,12 @@ Pod::Spec.new do |spec|
     ss.dependency 'OpenSSL-Static', '1.0.2.c1'
     ss.compiler_flags = folly_compiler_flags
     ss.source_files = 'iOS/SonarKit/FBDefines/*.{h,cpp,m,mm}', 'iOS/SonarKit/CppBridge/*.{h,mm}', 'iOS/SonarKit/FBCxxUtils/*.{h,mm}', 'iOS/SonarKit/Utilities/**/*.{h,m}', 'iOS/SonarKit/*.{h,m,mm}'
-    ss.public_header_files = 'iOS/SonarKit/**/{SonarClient,SonarPlugin,SonarConnection,SonarResponder,SKMacros}.h'
+    # Including FBMacros here is hacky since it causes cocoapods to copy
+    # FBMacros.h twice, but it's necessary since the "FBMacros.h" is always
+    # included in SonarKit-umbrella.h, but the actual path is "FBDefines/FBMacros.h",
+    # as defined by the FBDefines subspec. Copying the file an extra time allows
+    # the path in the umbrella header to resolve correctly.
+    ss.public_header_files = 'iOS/FBDefines/FBMacros.h', 'iOS/SonarKit/**/{SonarClient,SonarPlugin,SonarConnection,SonarResponder,SKMacros}.h'
```
priteshrnandgaonkar: Can't we get around this hack?

noahsark769: If we leave FBMacros.h in its current state, then I don't think we can simultaneously support the import in SKMacros.h that looks like this:

```
#import <FBDefines/FBMacros.h>
```

at the same time as the import that is generated in the umbrella header that looks like this:

```
#import "FBMacros.h"
```

We also can't remove the umbrella header import, as far as I know, because FBMacros.h needs to be a public_header_file of the subspec since otherwise it won't be able to be found when importing SKMacros.h.

I was thinking that since FBMacros.h seems to only be used in SKMacros.h, we could simply move the contents of it into SKMacros.h - @priteshrnandgaonkar would this break anything internally?

priteshrnandgaonkar: You can move the contents to `SKMacros.h`, it shouldn't cause internal things to break.

noahsark769:
@priteshrnandgaonkar We moved FBMacros into SKMacros, so we don't need the Podspec hack anymore :)

PR #207, opened 2018-08-07
Remove an extraneous comments from the iOS setup docs

noahsark769:
This comment was left there from the default cocoapods-generated Podfile, so it can now be removed.


passy:
Good catch. Thanks!

Issue #101, opened 2018-06-25
Openssl implementations creating .srl in wrong location (Permission Denied)


linseanyu:
I can't see Layout Inspector and Network Inspector in Sonar Mac app even when I run (Sonar/iOS/Sample)this sample project .I can only see Logs on it . 
![2018-06-25 8 23 35](https://user-images.githubusercontent.com/40568003/41850010-ab677b68-78b5-11e8-9c68-048c38acc827.png)

priteshrnandgaonkar:
bytw did u build sonar desktop app from the repo or downloaded from fbsonar.com?

linseanyu:
@priteshrnandgaonkar Thank you so much for helping . I've tried them both. But it's the same. I can only see Logs plugin there

noahsark769:
I also can't see the layout or network plugin in the Sonar desktop app (this is the case for master and for the download from fbsonar.com)

danielbuechele:
Sonar is currently not working with physical iOS devices. Simulator devices should work. Are you trying  to test it with our sample app or integrate it into your own apps?
Can you open the dev tools in Sonar (CMD+Option+I) and post any errors that are logged to the console?

noahsark769:
@danielbuechele this is with the sample app. There doesn't seem to be any info in the console, other than the fact that some Android things are failing (I think is because I don't have the Android SDK installed):

![image](https://user-images.githubusercontent.com/1168853/42129956-d1dc009e-7c88-11e8-9532-0d974c067128.png)

noahsark769:
Not sure what happened, but I can see the layout and network inspectors now. @linseanyu can you reproduce this anymore?

mengxiangyue:
I also cant see them, but i get this error.
![image](https://user-images.githubusercontent.com/1563329/42610287-64c5c56e-85c3-11e8-9b46-8822a59169e3.png)

erawhctim:
I'm seeing the same behavior on my end: 
<img width="539" alt="screen shot 2018-07-14 at 4 15 33 pm" src="https://user-images.githubusercontent.com/6600919/42728126-36e2963a-8781-11e8-9019-11ce94b8be64.png">

It looks like some permissions issue related to SSL? any thoughts?

jknoxville:
It looks like it's unable to open the serial file used for generating certificates, though it shouldn't really be creating that in your home directory.

[...]

PR #227, opened 2018-08-14
Explicitly specify serial filename to openssl

erawhctim:
Manually add `serial.srl` as the `CAserial` param to `openssl`. This
fixes a permissions issue when generating certificates, which fixes some
of the issues explained in #101.

### DIFF
--- a/src/utils/CertificateProvider.js
+++ b/src/utils/CertificateProvider.js
@@ -601,7 +601,6 @@ export default class CertificateProvider {
       .then(_ =>
         openssl('req', {
           new: true,
-          CAserial: getFilePath('server.srl'),
           key: serverKey,
           out: serverCsr,
           subj: serverSubject,
--- a/src/utils/CertificateProvider.js
+++ b/src/utils/CertificateProvider.js
@@ -27,6 +27,7 @@ const caKey = getFilePath('ca.key');
 const caCert = getFilePath('ca.crt');
 const serverKey = getFilePath('server.key');
 const serverCsr = getFilePath('server.csr');
+const serverSrl = getFilePath('server.srl');
 const serverCert = getFilePath('server.crt');
 
 // Device file paths
@@ -168,6 +169,7 @@ export default class CertificateProvider {
         CA: caCert,
         CAkey: caKey,
         CAcreateserial: true,
+        CAserial: serverSrl,
       });
     });
   }
@@ -599,6 +601,7 @@ export default class CertificateProvider {
       .then(_ =>
         openssl('req', {
           new: true,
+          CAserial: getFilePath('server.srl'),
           key: serverKey,
           out: serverCsr,
           subj: serverSubject,
@@ -611,6 +614,7 @@ export default class CertificateProvider {
           CA: caCert,
           CAkey: caKey,
           CAcreateserial: true,
+          CAserial: serverSrl,
           out: serverCert,
         }),
       )


jknoxville:
Hey @erawhctim I know I originally suggested using this, but after testing this it and it turns out it doesn't work with LibreSSL. I expected LibreSSL to conform to the openssl interface but this isn't the case for -CASerial: `unknown option -CAserial`

So we can't actually ship this as is. At the moment we don't have any logic to detect the openssl implementation and do something differently, and if possible it would be good to avoid that, though it is an option.

The strange thing is that the openssl docs say that the default location of the serial file used is the same directory as where the cert will go: https://www.openssl.org/docs/man1.0.2/apps/x509.html so it shouldn't actually be necessary at all.

But in your case that isn't happening. I wonder if you have any openssl.conf file that overrides this or something. Another option for you would be to make sure that you have r+w permissions on that file in `~/`

erawhctim:
@jknoxville Interesting! Thanks for the info. I can poke around and see, though I'm not very familiar with openssl at all so I'm not sure what I'd be looking for in a config file.

erawhctim:
@jknoxville That makes sense (saw the other issue that got opened). I can update this PR and take a stab at creating a way to check the openssl impl. from `openssl version`; I'm not sure of a better way to check for that sort of thing..

erawhctim:
@jknoxville Do you guys have a README/guide or any tips on getting an IDE setup to build/debug Sonar locally? I'm working through the changes discussed above (and currently debugging in the Chrome web tools within Electron) but its a pretty slow/manual process

jknoxville:
Hey @erawhctim that's a good point, I think all we have is this at the moment: https://fbflipper.com/docs/js-setup.html#working-on-the-core

I can write some docs up if you find this helpful. Below is my workflow:

I use Nuclide, which has built-in flow support, however this has recently been retired from open source. I haven't tried it with VSCode or Sublime but I would imagine they should be fine too. I highly recommend using an editor with flow support to highlight type errors etc without having to run the app.

You can also manually run `flow` or `yarn run flow` to type check everything, if for example you change an interface.

It should work with source maps enabled in electron, so you can at least use the un-minified code in the chrome debugger. Whenever I need to use breakpoints and step through code I use chrome dev tools like this, which works fine as far as I'm concerned.

I just tired out of curiosity and was able to connect Nuclide's debugger to Node by just entering port 9222, but it was using the bundled file rather than the original source files, not sure how easy that is to fix.

Typically I run yarn start, leave the app running, and the app should automatically refresh as you save JS files. If it doesn't, Cmd-R will refresh electron and load the latest code quickly - you don't need to close and reopen the app.

For completeness, we also have `yarn test` and `yarn test-electron` to run the tests.

Happy to answer anything more specific, I'm not sure which parts you're finding slow/manual.

@danielbuechele may have some more tips.

erawhctim:
@jknoxville that's incredibly helpful, thanks so much. I'll give this a try and report back with specific issues if I come across them

erawhctim:
@jknoxville making some progress now. I'm able to detect the openssl implementation via `openssl version`, but I also revisited my original commit and modified it slightly. 

I originally was manually specifying the `CAserial` param for the openssl `req` command, but I've updated it to only do that for the openssl `x509` commands, which _appears_ to be working perfectly (for the LibreSSL implementation). 

That leads me to two questions:
- Do we still need to conditionally add that param based on openssl implementation?
- What other implementations do we need to support, and how can we install and verify those?

src/utils/CertificateProvider.js
```diff
@@ -599,6 +601,7 @@ export default class CertificateProvider {
       .then(_ =>
         openssl('req', {
           new: true,
+          CAserial: getFilePath('server.srl'),
```
jknoxville: I think you're right that we just needed to take this arg out of the 'req' calls, and leave it in the 'x509' ones.
This one appears to be still left, but when I remove it, everything is fine - I don't think there were any differences in expected parameters between versions after all.

If you can confirm that with this change, it still works for you then we'll be good to go!
To do that, the easiest way is to wipe `~/.flipper`, which will force another cert exchange.

```suggestion
```

jknoxville:
That's great news @erawhctim , I've put a comment on the PR, I think maybe you forgot to include the change. Looks like we'll finally get this resolved!

Since I think it was us making the mistake of thinking the different implementations take different parameters, I don't think we'll need any version switch, and I also don't think we need to test on other versions, since the main docs specify this parameter: https://www.openssl.org/docs/man1.0.2/man1/x509.html

src/utils/CertificateProvider.js
erawhctim: Ah, great catch - left this in from the first time. Thanks!
Removing that doesn't change the behavior, seems to work expected 👍

erawhctim:
@jknoxville Agreed - this should put this issue to rest (finally 🙌). 

Thanks again for your help! Now I can finally get moving on my Realm DB plugin.

jknoxville:
Awesome, were always keen to hear about externally developed plugins, feel free to reach out if you need any guidance!

erawhctim:
Will do - I'm sure there are other issues and questions I'll come up with along the way.

@jknoxville For comments and questions that aren't necessarily _directly_ related to isssues/bugs with Flipper, do you have a preferred method of communication? Twitter page? New github issues with a specific label/naming convention? Public slack/discord/etc? Would love to pick your brain (and others on the team) with questions without being a PITA.

jknoxville:
Thanks for asking @erawhctim 
For quick questions, we have a twitter page you can use: https://twitter.com/flipper_fb
For discussions or more involved questions then filing an issue with the question label is probably the way to go.

PR #238, opened 2018-08-16
Use UUID instead of new Random().nextInt(...)

yschimke:
Minor simplification, mostly cleanup.  Arguably, better to avoid the birthday paradox e.g. collision with > a few thousand requests. 

Tested via 

$ ./gradlew :sample:installDebug


passy:
That's awesome. Thanks, Yuri!

PR #239, opened 2018-08-16
Allow duplicate headers for OkHttp

yschimke:
The Android side of supporting multiple headers.


yschimke:
@passy I'm struggling with the change needed in src/plugins/network/RequestDetails.js, around HeaderInspector.render, can't assume header key is unique

passy:
Cleaning up old issues and pull requests. Please let us know if this is something that's still relevant and we can reopen this and take another look. Thanks!

PR #275, opened 2018-09-19
[desktop] Fix initialization

### LINKED ISSUE #267: Desktop app appear bug.
I download the latest build desktop app and open it appear this bug.
default

passy:
Summary:
The global variable generated by metro(?) was changed from Sonar to
Flipper without updating its callside.

Fixes #267

Test Plan:
Break point at `index.dev.html:89 Uncaught TypeError: Cannot read
property 'init' of undefined` shows that Sonar is undefined while
Flipper is. Manually calling `Flipper.init()` unbreaks it.

### DIFF
--- a/static/index.dev.html
+++ b/static/index.dev.html
@@ -86,7 +86,7 @@
           };
 
           script.onload = () => {
-            global.Sonar.init();
+            global.Flipper.init();
           };
 
           document.body.appendChild(script);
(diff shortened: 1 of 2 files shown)


passy:
@danielbuechele is already landing a fix.

carlonzo:
@passy @danielbuechele when can we have a release with the fix? we are currently blocked by this!
thanks for your help! :)

passy:
@carlonzo Sorry, we're still in the middle of landing some rename changes and would like to have them all in place before the next release as this introduces a ton of breaking changes. It'll probably take a few more days into the next week before we're ready.

Issue #145, opened 2018-07-08
UserDefaults editor plugin for iOS?


noahsark769:
A SharedPreferences editor for Android was just open sourced. Are there plans to extend it to work with iOS UserDefaults as well? Or would this be something the community could build?

priteshrnandgaonkar:
There are no plans yet for iOS from us. But we would be happy to have a PR from open source community.

hhtopcu:
I'm planning to work on it. I'll send a PR asap

pkrmf:
@priteshrnandgaonkar I have an [PR open](https://github.com/facebook/flipper/pull/291) 

Want to add a few things to it, but feel free to start reviewing it.

priteshrnandgaonkar:
Thanks for the PR, I will look into it.

priteshrnandgaonkar:
We support it now, check the sample app

PR #291, opened 2018-10-07
Integrating NSUserDefaults plugin to iOS

pkrmf:
I have a few details left, but its almost done. This PR addresses #145

- The NSUserDefaults plugin uses the SharedPreferences Desktop Part since we can reuse all of it.
- The NSUserDefaults plugin uses swizzling in order to be notified of what specific event changed at runtime.
- Added Test harness in both Sample Swift and Sample apps for iOS in order to test the plugin.
- Updated the documentation in `docs/shared-preferences-plugin.md` and` README.md`

I am open to suggestions since the desktop sharedPreferences version doesn't support deletion of preferences. Most likely I would have to modify the UI, and for that matter, I might as well build a user defaults desktop version

I wanted to add @xiphirx in this MR since he developed the shared preferences plugin for Android and Desktop. I don't see a way to remove preferences from the flipper desktop app so I was wondering if you would be OK with me adding that.


pkrmf:
How do we go about versioning?

xiphirx:
Adding the ability to delete keys in the shared preferences plugin sounds good to me.

iOS/Plugins/FlipperKitUserDefaultsPlugin/FKUserDefaultsPlugin.m
```diff
@@ -0,0 +1,79 @@
...
+    if (self = [super init]) {
+        _userDefaults = [NSUserDefaults standardUserDefaults];
+        _suiteName = suiteName;
+        [FKUserDefaultsSwizzleUtility swizzleSelector:@selector(setObject:forKey:) class:[NSUserDefaults class] block:^(NSInvocation * _Nonnull invocation) {
+            __unsafe_unretained id firstArg = nil;
+            __unsafe_unretained id secondArg = nil;
+            [invocation getArgument:&firstArg atIndex:2];
+            [invocation getArgument:&secondArg atIndex:3];
+            [invocation invoke];
+            [self userDefaultsChangedWithValue:firstArg key:secondArg];
```
priteshrnandgaonkar: `weak self ` here, since this block is stored in a static variable.

iOS/Plugins/FlipperKitUserDefaultsPlugin/FKUserDefaultsSwizzleUtility.m
```diff
@@ -0,0 +1,59 @@
...
+//  FlipperKit
+//
+//  Created by Marc Terns on 10/6/18.
+//
+
+#import "FKUserDefaultsSwizzleUtility.h"
+#import <objc/runtime.h>
+
+static NSMutableSet *swizzledClasses;
+static NSMutableDictionary *swizzledBlocks;
```
priteshrnandgaonkar: Can we get rid off this static variables ?

pkrmf: I can build a singleton and declare the properties as strong/copy. Then I can have a private instance method to do the swizzling. The swizzleutility needs to be a singleton since it keeps references of all the blocks to be called once the swizzling happens and we can not lose those references whatsoever(they need to be alive as long as the process is alive). 

I can add also comments to every single line of the FKUserDefaultsSwizzleUtility to better explain what it does.

pkrmf: Hey @priteshrnandgaonkar, I made the changes as requested in 12ec6096af378f4e09d92c18381f23215f43ffdf

pkrmf:
@priteshrnandgaonkar This PR should be ready

priteshrnandgaonkar:
Importing it in our system now, thanks for working on it.

PR #309, opened 2018-10-26
Make guava dependency test-only

### LINKED ISSUE #172: Remove Guava dep
It seems like we're pulling in the full Guava library for just `ImmutableMap`. That's overkill to put it mildly. We should either find a smaller implementation of that somewhere out there or provide one ourselves.

hzsweers:
It's only used in `JavascriptSessionTest`, and seems fine to keep it in tests

Resolves #172

### DIFF
--- a/android/src/test/java/com/facebook/flipper/plugins/console/JavascriptSessionTest.java
+++ b/android/src/test/java/com/facebook/flipper/plugins/console/JavascriptSessionTest.java
@@ -9,9 +9,9 @@ package com.facebook.flipper.plugins.console;
 
 import static org.junit.Assert.assertEquals;
 
-import com.google.common.collect.ImmutableMap;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Map;
 import org.json.JSONObject;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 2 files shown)


hzsweers:
I would look into it but I can’t index the project as it stands in master because not all the Gradle plugins resolve

jknoxville:
Thanks @hzsweers that's definitely an improvement but I think Hilal has a good point that we really don't need this dep at all. I'd rather we just remove it and use HashMap in that test instead.

hzsweers:
Sure, I'll update it

hzsweers:
Removed. Can't run locally to verify it works though because master appears to be red. Note that guava is still on the test classpath from various test libraries

jknoxville:
Thanks, I'll get this merged asap.
Sorry about the travis failure, we've got a fix for that in the works too. However the reason that is failing is because travis runs the build in a single thread, which recent releases of metro couldn't deal with. But on any "normal" machine, it should work.

Feel free to report a new issue if you're having trouble building locally.

android/src/test/java/com/facebook/flipper/plugins/console/JavascriptSessionTest.java
```diff
@@ -42,12 +42,13 @@ public void testStatePersistsBetweenCommands() throws Exception {
 
   @Test
   public void testVariablesGetBoundCorrectly() throws Exception {
+    Map<String, Object> data = new HashMap<>();
```
pavlospt: you could also make it `final` if you want :)

passy:
Thanks for the fix, @hzsweers! Much appreciated!

hzsweers:
Happy to help!

PR #313, opened 2018-10-31
Fix OkHttp Interceptor closing response on IOException.

NightlyNexus:
The returned response body should not be closed, so that users can read the response body.

Before, when [this call](https://github.com/facebook/flipper/blob/db833d36e9e318252e01f9a7d4608702634e673c/android/src/main/java/com/facebook/flipper/plugins/network/FlipperOkhttpInterceptor.java#L94) threw an exception, the body would be closed, and then [this](https://github.com/facebook/flipper/blob/db833d36e9e318252e01f9a7d4608702634e673c/android/src/main/java/com/facebook/flipper/plugins/network/FlipperOkhttpInterceptor.java#L54) would return a closed response.


passy:
Ah, I've heard about this from others before. Thanks a lot for fixing this!

PR #321, opened 2018-11-07
Fix a couple of Sonar references in FlipperKit.podspec

noahsark769:
Noticed these while looking through the podspec.


jknoxville:
Cheers Noah! 👍

PR #322, opened 2018-11-11
[iOS]Crash reporting plugin

priteshrnandgaonkar:
A barebone of a crash reporting plugin. It catches the crash which may happen in iOS app and would send the crash stack as a message to the desktop app.

- [x] Notifications for crash

A screenshot of the message looks like the following
screen shot 2018-11-12 at 3 02 11 am


iOS/Sample/NetworkViewController.m
```diff
@@ -17,6 +17,8 @@ @implementation NetworkViewController
 - (void)viewDidLoad {
   [super viewDidLoad];
   self.navigationItem.title = @"Network";
+  NSArray *array = @[];
+  [array objectAtIndex:4];
```
priteshrnandgaonkar: This is added deliberately to cause a crash. I will remove it once I am done with my final changes.

priteshrnandgaonkar:
It got merged

Issue #341, opened 2018-12-14
iOS Swift Integration - Search Paths and Flags not getting setup


dmac81:
After following the documented setup steps on a single page swift application and setting up the pods, after importing FlipperKit and then initialize FlipperClient: `let client = FlipperClient()` Xcode will not be able to find FlipperClient, after searching high and low and comparing to the swift sample project I found that the below Xcode settings are set in your sample project but integrating your Pod does not cause these to be added or noted that they need to be added

HEADER_SEARCH_PATHS, OTHER_CPLUSPLUSFLAGS, OTHER_LDFLAGS, & OTHER_SWIFT_FLAGS

Am I missing something?

priteshrnandgaonkar:
HEADER_SEARCH_PATHS should match with sample app. I just checked that the cocoapod doesn't automatically add `-DFB_SONARKIT_ENABLED` flag which is required in OTHER_CPLUSPLUSFLAGS, OTHER_LDFLAGS, & OTHER_SWIFT_FLAGS. I will look into it and update the podspec. Thanks for reporting the issue.

dmac81:
Also to get these pods to work in a project that use dynamic frameworks, here is some code to force Flipper to stay static

```
$static_framework = ['FlipperKit', 'Flipper', 'Folly', 
    'CocoaAsyncSocket', 'ComponentKit', 'DoubleConversion', 
    'glog', 'PeerTalk', 'RSocket', 'Yoga', 'YogaKit', 
    'CocoaLibEvent', 'OpenSSL-Static', 'boost-for-react-native']
```

```
pre_install do |installer|
    installer.pod_targets.each do |pod|
        if $static_framework.include?(pod.name)
            pod.host_requires_frameworks = false
        end
    end
end
```

priteshrnandgaonkar:
Closing the issue as it is resolved.

PR #343, opened 2018-12-18
[iOS] Docs with an updated installation instruction

priteshrnandgaonkar:
This PR updates the installation instruction. It missed few crucial information and thus the projects installing flipper as a dependency were facing issues. It addresses the issue #341. It also brings the sample application at par with the documentation.


dmac81:
I will post a follow up to this once it's merged to include how to mix dynamic and static libraries in CocoaPods to help those that have to use `use_frameworks!`

PR #352, opened 2019-01-06
Enable override of context menu on ManagedTable.

paulovap:
Summary: This patch add a new prop on ManageTabled called buildContextMenuItems. It allows
to override the current context menu when right click on the ManagedTable rows.


passy:
Thanks for the contribution, @paulovap!

PR #354, opened 2019-01-15
Fix soloader gradle dependency

Arne517:
The gradle dependency on soloader was using a wrong groupId (flipper instead of soloader)


passy:
Woops, thanks for catching that. Looks like one of our scripts was a bit overeager.

PR #355, opened 2019-01-16
[iOS] Update folly to v2019.01.14.00

priteshrnandgaonkar:
The folly version which we were using to build open source was v2018.08.06.00, which is very outdated.

This PR updates the version of the folly to the latest one.


priteshrnandgaonkar:
Reason is that the sample app is trying to pull the Folly dependency from github master branch, which obviously doesn't have Folly version 1.2.0. I will update the way sample app references its dependencies.

PR #357, opened 2019-01-21
Upgrade to Gradle 5.1.1 stable version

pavlospt:
_Per title_


gradlew.bat
```diff
@@ -0,0 +1,84 @@
+@if "%DEBUG%" == "" @echo off
```
passy: I'd actually rather leave this one out to not imply we support building on Windows (which we don't) :D

pavlospt: @passy `gradlew.bat` has been removed :)

PR #362, opened 2019-01-22
Properly fix license issue in CircleCI

pavlospt:
Apparently this seems to be the only way to make build work, with the updated licenses. It also follows the same strategy as Travis.

I am using double pipe in `sdkmanager` command execution, to avoid breaking the build, because `sdkmanager` returns exit code 141 in case it meets the requirements requested. (e.g.: `yes | sdkmanager --licenses` returns 141, which in turn breaks the build due to `set -eo pipefail` Bash flags, that CircleCI is using in order to execute any scripts)

For the record, I tried the following solutions, that did not work:
* Add accepted license SHA in Android SDK licenses folder.
* Accept all licenses prior to running the other jobs (with a separate job, working as a dependency to the rest of the Workflow jobs) (`yes | sdkmanager --licenses`)
* Accept all licenses prior to executing any Gradle command (without a separate job) (`yes | sdkmanager --licenses`)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,17 +9,6 @@ executors:
       TERM: 'dumb'
 
 jobs:
-  add_accepted_android_sdk_license:
-    executor: default-executor
-    docker:
-      - image: circleci/android:api-28-ndk-r17b
-    steps:
-      - run:
-          name: Add accepted Android SDK license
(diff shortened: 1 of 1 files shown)


passy:
Thank you, Pavlos! :)

Issue #132, opened 2018-07-04
Publish SonarKit.podspec to the CocoaPods Master Repository


pkrmf:
Currently SonarKit is published to the Facebook Sonar private repository. In order to get SonarKit.podspec published to the CocoaPods master repo, we first need to take care of fixing and publishing some of its direct dependencies:

- [x] Yoga podspec should be updated to include the `DEFINES_MODULE` directive. We should be opening an issue in [Yoga](https://github.com/facebook/yoga), a PR and ask the owners of the repository to pod trunk the new podspec. There are other workarounds to avoid modifying the podspec, like setting the `:modular_headers` directive for `Yoga` in a Podfile, but that would prevent us from being able to lint and push the SonarKit.podspec to CocoaPods.

- [x] YogaKit has a dependency on the patch version of Yoga. We should tell the repository owners to avoid that, otherwise, if Yoga gets updated to 1.9 due to new API , or changes on its podspec, YogaKit won't be able to pull a new minor version due to directly defining a dependency on a patch version. Long story short, YogaKit should depend on Yoga ~>1.8 instead of Yoga ~>1.8.1 or CocoaPods will treat 1.9 for example as a major version change, when it is not. Let's make sure they publish the new podspec to the cocoapods master repo.

- [ ] Sonar. Well, unfortunately for Sonar.podspec, there is already a [published pod with the same name](https://github.com/CocoaPods/Specs/blob/master/Specs/0/c/a/Sonar/3.0.0/Sonar.podspec.json). There is no much we can do here, unless we change Sonar podspec name to something else so we can publish it, or build a subspec in SonarKit where we define Sonar, or make Sonar subspec part of our Core subspec. If we don't see other iOS developers consuming Sonar.podspec as a direct dependency, and its always through SonarKit, maybe a subspec would be the best option here.

- [x] RSocket.podspec needs to be published to the cocoapods master repo. It doesn't exist yet, so that is good for us. We need to create an issue and a PR in [RSocket repository](https://github.com/rsocket/rsocket-cpp)

- [ ] Peertalk needs to also be updated. I already opened an issue [#48](https://github.com/rsms/peertalk/issues/48) in PeerTalk asking them to generate a new tag using the latest commit since their podspec is pointing to an old commit that doesn't include the latest codebase.

[...]

PR #395, opened 2019-03-12
Published Flipper-PeerTalk dependency on Cocoapods

priteshrnandgaonkar:
I have published the new tag of Peertalk on my [fork](https://github.com/priteshrnandgaonkar/peertalk) and published a [Flipper-PeerTalk](https://github.com/CocoaPods/Specs/blob/80e4cd9a05f45579a82a41c96afe3e8af3b2c08f/Specs/d/6/6/Flipper-PeerTalk/0.0.4/Flipper-PeerTalk.podspec.json) pod on cocoapods. There was no reply on the [request](https://github.com/rsms/peertalk/issues/48) to publish the new version of the peertalk on their repository and it was blocking us to publish Flipper and FlipperKit pods to Cocoapods. For more context follow the issue #132. With this PR we would have all our dependencies on cocoapods and thus we can now publish Flipper and FlipperKit pods on the cocoapods master repository. 

I have also added the Flipper-PeerTalk podspec file for future updates.


pkrmf:
This is excellent work! Thabks for finishing it, I hit those roadblocks a long time ago and I wasn’t sure how to proceed.

Is it possible to update documentation, sample apps and remove the private specs repo as psrt of this PR?

priteshrnandgaonkar:
I will update documentation and sample app as a separate PR. I will still keep the private specs in the `Specs` folder, as some users still might be using the older version of Flipper. I don't want to break for them.

PR #397, opened 2019-03-17
Fix `DataInspector` Component `diff` some problems.

lijy91:
1. In `diffMetadataExtractor` function, when the `data` or `diff` is `null`, it diff view does not display correctly.

```jsx

<Panel floating={false} heading={'Test'}>
  <ManagedDataInspector
    diff={{
      auth: null
    }}
    data={{
      auth: {
        user: {
          name: 'JianyingLi'
        }
      }
    }}
    expandRoot={true}
  />
</Panel>
```

Before: 
2561552808023_ pic

After: 
2551552807969_ pic

2. Status `added`(green) and `removed`(red) should be the opposite

```jsx
<ManagedDataInspector
  diff={{
    user: {
      name: 'Leo'
    }
  }}
  data={{
    user: {
      name: 'JianyingLi'
    }
  }}
  expandRoot={true}
/>
```

Before: 
2591552808195_ pic

After: 
2581552808145_ pic


passy:
Thanks for this and the great summary here, @lijy91!

PR #400, opened 2019-03-21
`iOS Devices are not yet supported` Displayed only when iOS physical device is selected

lijy91:


lijy91:
@passy can you review this?

src/plugins/network/RequestDetails.js
```diff
@@ -532,6 +532,7 @@ class JSONTextFormatter {
   format = (body: string, contentType: string) => {
     if (
       contentType.startsWith('application/json') ||
+      contentType.startsWith('application/hal+json') ||
```
passy: This stuff here seems unrelated. Could you split this up into a separate PR? :)

lijy91: I split into two pr, please review again

PR #404, opened 2019-03-31
Fix PR #400 Unable to set error message incorrectly

lijy91:
This is a bug in #400


jknoxville:
I think what you're trying to do is make the error disappear when you deselect the unsupported iOS device.

In that case, if you check the value of the existing error, we can safely clear it if it was this iOS error, and leave it unchanged if not.

lijy91:
@jknoxville Please review again.

PR #409, opened 2019-04-05
Create DEV_ENVIRONMENT.md

jknoxville:
Adding a recommended workflow for those that don't have a good setup after conversation in https://github.com/facebook/flipper/pull/227

Just a starting point but I think it's better than nothing.

### DIFF
--- a/DEV_ENVIRONMENT.md
+++ b/DEV_ENVIRONMENT.md
@@ -1,0 +1,32 @@
+# Developer Environment Setup
+
+## Desktop app and Flipper Plugins (Javascript)
+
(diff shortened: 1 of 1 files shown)


DEV_ENVIRONMENT.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Editor**: Android Studio
+
+## iOS plugins and SDK (Objective-C)
+
+**Editor**: XCode
+
+## Cross-platform SDK (Used by Android and iOS SDK) (C++)
+
+**Editor**: XCode, Visual Studio Code, Vim...
```
passy: I think Android Studio is actually not terrible here?

jknoxville: Added.

Issue #408, opened 2019-04-05
New activities not automatically added to the Layout hierarchy tree


qichuan:
![Capto_Capture 2019-04-05_03-20-55_PM](https://user-images.githubusercontent.com/410850/55611149-4634ae00-57b7-11e9-8cef-94af91a0af12.gif)

As shown in the video, the user needs to exit the Layout page and re-enter it in order to see the new Activity (`FlipperDiagnosticActivity` in this case) being added to the hierarchy tree

jknoxville:
Thanks for the report, looks like a bug.

qichuan:
I think I know how to fix it, will create a PR later.

mgoyanes:
@qichuan What's that images plugin?

PR #411, opened 2019-04-05
Fix issue 408

qichuan:
Fix https://github.com/facebook/flipper/issues/408

### DIFF
--- a/src/plugins/layout/Inspector.js
+++ b/src/plugins/layout/Inspector.js
@@ -87,8 +87,8 @@ export default class Inspector extends Component<Props> {
           .reduce((acc, cv) => acc.concat(cv), []);
         this.getNodes(ids, {}).then(nodes => {
           nodes.forEach(node => {
-            // Check if there are new IDs in the children array, and call getNodes() 
-            // if there are any 
+            // Check if there are new IDs in the children array, and call getNodes()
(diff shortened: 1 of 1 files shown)


jknoxville:
Awesome, thanks for the quick submission! Looks good on the surface, but I'll review it in more depth.

jknoxville:
Hey @qichuan the reason I haven't merged this yet is I think it will cause the initial getNodes calls to recurse all the way down the tree and end up getting every node in the view hierarchy (getNode calls updateElements, and now updateElements would call getNode on the children). At the moment, we do this lazily when each node gets expanded.
If that's true then I think a better approach would be to not add any new recursion here, but to make sure the children are updated at the top level, e.g. calling getChildren inside the INVALIDATE handler.

qichuan:
@jknoxville Thanks for the finding, let me think about it and come back to you soon.

qichuan:
@jknoxville So I have updated the PR with your suggestion, please take a look.

In addition, I found that the iOS Layout plugin has a similar issue, the new ViewControllers are not automatically added to the view tree, and it turns out that the invalidate message is never sent. I will create another issue for this.

jknoxville:
Great! Many thanks @qichuan , merging it now.

PR #414, opened 2019-04-15
Bump some more dependencies

passy:
Summary: Some more minor upgrades.

Differential Revision: D14933687

### DIFF
--- a/android/sample/build.gradle
+++ b/android/sample/build.gradle
@@ -13,10 +13,16 @@ android {
     lintOptions {
         abortOnError false
     }
+
     sourceSets {
         main {
             manifest.srcFile './AndroidManifest.xml'
         }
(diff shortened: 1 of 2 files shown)


android/sample/build.gradle
```diff
@@ -13,11 +13,17 @@ android {
...
     }
+
     sourceSets {
         main {
             manifest.srcFile './AndroidManifest.xml'
         }
     }
+
+    compileOptions {
+        targetCompatibility = "8"
```
pavlospt: @passy You can use `JavaVersion.VERSION_1_8` as an alternative for this kind of Java version constants :)

passy: Oh, good idea. Thanks!

PR #415, opened 2019-04-17
Add copy as cURL for network requests

### LINKED ISSUE #406: Copy as cURL and other network options
Would be great to implement stetho style 'Copy as cURL' and other options

image

bdotdub:
Summary:
This diff adds a context menu item for network request rows that allows the user to copy the request as a curl command. The logic for the command generation was inspired by [`FLEXNetworkCurlLogger`](https://github.com/Flipboard/FLEX/blob/master/Classes/Network/FLEXNetworkCurlLogger.m)

image

Resolves #406

Differential Revision: D14973754

### DIFF
--- a/src/plugins/network/__tests__/requestToCurlCommand.node.js
+++ b/src/plugins/network/__tests__/requestToCurlCommand.node.js
@@ -1,0 +1,145 @@
+/**
+ * Copyright 2018-present Facebook.
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ * @format
+ */
+
+import {convertRequestToCurlCommand} from '../utils.js';
+import type {Request} from '../types.js';
+
+test('convertRequestToCurlCommand: simple GET', () => {
+  const request: Request = {
+    id: 'request id',
+    timestamp: 1234567890,
+    method: 'GET',
+    url: 'https://fbflipper.com/',
+    headers: [],
+    data: null,
+  };
+
+  const command = convertRequestToCurlCommand(request);
+  expect(command).toEqual("curl -v -X GET 'https://fbflipper.com/'");
+});
+
+test('convertRequestToCurlCommand: simple POST', () => {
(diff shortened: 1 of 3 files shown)


src/plugins/network/index.js
```diff
@@ -192,6 +232,29 @@ export default class extends FlipperPlugin<State, *, PersistedState> {
...
+    const {requests} = this.props.persistedState;
+    const {selectedIds} = this.state;
+    // Ensure there is only one row highlighted.
+    if (selectedIds.length !== 1) {
+      return;
+    }
+
+    const request = requests[selectedIds[0]];
+    let command: string = `curl -v -X ${request.method}`;
+    command += ` "${request.url}"`;
```
passy: Do we need to apply some extra escaping here (and in the other places we insert user data)? It seems like it would be feasible to create a malicious cURL command by adding special shell characters here, e.g. `; rm -rf` or `$(cat /etc/passwd | curl ...`.

bdotdub: Good call! [The Chrome inspector](https://github.com/ChromeDevTools/devtools-frontend/blob/a86c14edda076df6d273e97b61359e8e7fe8ff86/front_end/network/NetworkLogView.js#L1646-L1761) seems to do something. I'll try my hand at this

src/plugins/network/__tests__/requestToCurlCommand.node.js
```diff
@@ -0,0 +1,145 @@
...
+ * Copyright 2018-present Facebook.
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ * @format
+ */
+
+import {convertRequestToCurlCommand} from '../utils.js';
+import type {Request} from '../types.js';
+
+test('convertRequestToCurlCommand: simple GET', () => {
```
bdotdub: @passy Ready for your  👀 again! I added a few tests here to test the escaping

PR #420, opened 2019-04-29
React docgen update

passy:
Depends on #419.

Summary:
Major bump, but output looks like the same. Notably, it points out
that it can't parse ManagedTable.js because it gets hung up on some
Flow syntax, but that appears to already happen with 3.0, it's just that
the error messages improved.

Test Plan:
yarn start


passy:
Superseded.

PR #441, opened 2019-05-09
Add databases plugin v0 (android)

jknoxville:
## Summary

Adds a plugin for listing the databases, tables and contents of those tables in an android app.
Right now, works with sqlite, but it should be generic enough to work with other db types.

## Changelog

Add initial version of android databases plugin

Creating a PR, I may need to do some cleaning up, but this is to kick off that process.


android/no-op/.classpath
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
passy: I don't think we want any of these files.

android/sample/.settings/org.eclipse.buildship.core.prefs
```diff
@@ -0,0 +1,2 @@
+connection.project.dir=../..
```
passy: These can be removed, too.

android/sample/src/main/java/com/facebook/flipper/sample/FlipperSampleApplication.java
```diff
@@ -33,5 +37,126 @@ public void onCreate() {
...
+    DatabaseUtils.queryNumEntries(db1Helper.getReadableDatabase(), "db1_first_table", null, null);
+    DatabaseUtils.queryNumEntries(db2Helper.getReadableDatabase(), "db2_first_table", null, null);
+  }
+
+  public static class Database1Helper extends SQLiteOpenHelper {
+
+    // If you change the database schema, you must increment the database version.
+    public static final int DATABASE_VERSION = 4;
+    private static final String SQL_CREATE_FIRST_TABLE =
+        "CREATE TABLE " + "db1_first_table" + " (" +
```
passy: nit: Any reason for concatenating in the same line here?

android/sample/src/main/java/com/facebook/flipper/sample/FlipperSampleApplication.java
```diff
@@ -33,5 +37,126 @@ public void onCreate() {
...
+        "CREATE TABLE " + "db1_empty_table" + " (" +
+            "_id INTEGER PRIMARY KEY," +
+            "column1 TEXT," +
+            "column2 TEXT)";
+
+    public Database1Helper(Context context) {
+      super(context, "database1.db", null, DATABASE_VERSION);
+    }
+
+    public void onCreate(SQLiteDatabase db) {
```
passy: Are these `@Override`?

android/src/main/java/com/facebook/flipper/plugins/databases/DatabaseDescriptor.java
```diff
@@ -0,0 +1,5 @@
+package com.facebook.flipper.plugins.databases;
+
+public interface DatabaseDescriptor {
```
passy: Some docs would be great here.

android/src/main/java/com/facebook/flipper/plugins/databases/DatabaseDescriptor.java
```diff
@@ -0,0 +1,5 @@
+package com.facebook.flipper.plugins.databases;
```
passy: We need copyright headers on all of these. :)

android/src/main/java/com/facebook/flipper/plugins/databases/DatabaseDriver.java
```diff
@@ -0,0 +1,139 @@
...
+
+public abstract class DatabaseDriver<DESCRIPTOR extends DatabaseDescriptor> {
+
+    protected Context mContext;
+
+    public DatabaseDriver(@NonNull final Context context) {
+        mContext = context;
+    }
+
+    @NonNull
```
passy: We don't use `@NonNull`. This is our default; we use `@Nullable` instead.

android/src/main/java/com/facebook/flipper/plugins/databases/DatabaseDriver.java
```diff
@@ -0,0 +1,139 @@
...
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.StringDef;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.util.List;
+
+public abstract class DatabaseDriver<DESCRIPTOR extends DatabaseDescriptor> {
+
+    protected Context mContext;
```
passy: This should be private and final.

android/src/main/java/com/facebook/flipper/plugins/databases/DatabasesErrorCodes.java
```diff
@@ -0,0 +1,11 @@
+package com.facebook.flipper.plugins.databases;
+
+public class DatabasesErrorCodes {
```
passy: Seems to me like an enum would be a better fit for this.

android/src/main/java/com/facebook/flipper/plugins/databases/DatabasesFlipperPlugin.java
```diff
@@ -0,0 +1,50 @@
+// Copyright 2018-present Facebook. All Rights Reserved.
```
passy: That's not the right header here.

android/src/main/java/com/facebook/flipper/plugins/databases/DatabasesFlipperPlugin.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public void onDisconnect() {
+        databasesManager.setConnection(null);
+    }
+
+    @Override
+    public boolean runInBackground() {
+        return false;
```
passy: Just thinking out loud here, but it would be great if we could find a way to make this available "offline". Not sure how feasible that actually is given that databases could be quite large.

android/src/main/java/com/facebook/flipper/plugins/databases/DatabasesManager.java
```diff
@@ -0,0 +1,202 @@
...
+    private final static String GET_TABLE_DATA_COMMAND = "getTableData";
+    private final static String GET_TABLE_STRUCTURE_COMMAND = "getTableStructure";
+    private final static String EXECUTE_COMMAND = "execute";
+
+    private final List<DatabaseDriver> databaseDriverList;
+
+    private final SparseArray<DatabaseDescriptorHolder> databaseDescriptorHolderSparseArray;
+    private final Set<DatabaseDescriptorHolder> databaseDescriptorHolderSet;
+
+    private FlipperConnection connection;
```
passy: Naming scheme here doesn't match that of other files. Should be `mConnection`, `mDatabaseDriverList`, etc.

android/src/main/java/com/facebook/flipper/plugins/databases/DatabasesManager.java
```diff
@@ -0,0 +1,202 @@
...
+            @Override
+            public int compare(DatabaseDescriptorHolder o1, DatabaseDescriptorHolder o2) {
+                return o1.databaseDescriptor.name().compareTo(o2.databaseDescriptor.name());
+            }
+        });
+    }
+
+    public void setConnection(FlipperConnection connection) {
+        this.connection = connection;
+        if (connection != null) {
```
passy: Any reason for allowing empty connections here?

android/src/test/java/com/facebook/flipper/plugins/databases/DatabasesFlipperPluginTest.java
```diff
@@ -0,0 +1,532 @@
...
+import java.util.Arrays;
+import java.util.List;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
+
+@RunWith(RobolectricTestRunner.class)
+public class DatabasesFlipperPluginTest {
```
passy: Yay!

build.gradle
```diff
@@ -79,7 +79,7 @@ ext.deps = [
         lithoTesting       : "com.facebook.litho:litho-testing:$LITHO_VERSION",
         // Debugging and testing
         guava              : 'com.google.guava:guava:20.0',
-        robolectric        : 'org.robolectric:robolectric:3.3',
+        robolectric        : 'org.robolectric:robolectric:4.2',
```
passy: We can't do that. :(

This needs to match our internal version.

gradle.properties
```diff
@@ -27,3 +27,6 @@ org.gradle.internal.repository.initial.backoff=1250
 
 android.useAndroidX=true
 android.enableJetifier=true
+
+# Robolectric
+android.enableUnitTestBinaryResources=true
```
passy: I assume this won't work in 3.3?

android/sample/src/main/java/com/facebook/flipper/sample/FlipperSampleApplication.java
quanturium: Easier to read SQL code formatted this way

android/src/main/java/com/facebook/flipper/plugins/databases/DatabasesErrorCodes.java
quanturium: I'm going to keep it this way for now. We can change this in the future.

android/src/main/java/com/facebook/flipper/plugins/databases/DatabasesManager.java
quanturium: Yes, resetting the connection when onDisconnect() is called

build.gradle
quanturium: Understood

AnkitPatanaik:
I'm interested in using this plugin! How do I import this plugin (do I need to add anything to my gradle file?) into my Android app? @jknoxville

jknoxville:
Great to hear there's interest in it!

Take a look at the sample app for an example. I believe all you need to do
is initialise the plugin with the app context, and add it to the flipper
client just like any other plugin. I don't think you'll need to modify the
gradle files.

Let me know if you have issues - we only merged it today and you're the
first guinea pig to try it as far as I'm aware :) There may be some issues
to iron out.

AnkitPatanaik:
Thanks for the quick response @jknoxville! I've only started looking into Flipper and still kind of unfamiliar, but I think my issue is that I don't know how to install a plugin. Whenever I try to add a plugin, I'm unable to find the proper import for the plugin I tried to add to my FlipperClient

abaker:
@jknoxville It looks good to me so far, thanks! I would love to be able to execute SQL statements in Flipper

@AnkitPatanaik I just migrated to Flipper yesterday, you can look at this diff [here](https://github.com/tasks/tasks/commit/4997ec3f0e1f756fcad48b6a7eff90eff8c0e3ad)

AnkitPatanaik:
Thanks for the help @abaker! Looks like I got it up and running.

@jknoxville, great work! I would also like to echo @abaker's suggestion for adding the ability to execute SQL statements in Flipper.

Nimrodda:
@jknoxville I had a look into the code in the PR and I see that the infrastructure for executing custom queries is pretty much ready on the Android side, but on the Flipper UI side the button for executing queries is currently hidden and doesn't have the click handler. Is that still being worked on? I checked your fork and haven't seen new commits regarding that feature.

I'm considering making a PR to add this functionality and that's why I'm asking. I don't want to work on something that somebody else is already working on. Cheers!

jknoxville:
For those waiting for sql execution, [we now have this feature](https://github.com/facebook/flipper/issues/453), and it will be included in the next release 🚀

PR #545, opened 2019-09-11
linux instructions to build

a-c-sreedhar-reddy:


passy:
Thanks for raising this! We now offer prebuilt versions for Linux and Windows, too. For now, they're available directly on the release page, but we'll make sure to add them to the homepage as well shortly.

I'll add a note inspired by yours to the readme.

PR #563, opened 2019-09-27
Fix crash of iOS FlipperNetworkPlugin

LAgagggggg:
## Summary
  Flipper's network plugin may cause crash on a frequency of about one time per 1~2 days(I think it depends on how many network request we send).
IMG_3095

  I assume the crash is caused by part of the code of HTTPParser is not thread-safe, and invoke it asynchronously may cause double-free crash. So I manually ask for HTTPBody synchronously, make the HTTPParser parse the request and cache them in advance, before any possible asynchronous invoking.

## Changelog
  Fix potential crash cause by network plugin.

## Test Plan
1. Run our App with Flipper integrated
2. Make sure FlipperNetworkPlugin is active
3. It may take days to reproduce the crash


priteshrnandgaonkar:
Thanks for the PR. I will have a look in few days.

PR #568, opened 2019-10-03
corrected some spelling mistakes, some punctuations

lokeshgupta0912:


passy:
Hey, @lokeshgupta0912. Thanks for this, but it seems like this is just an automatic spell checker and all of the instances are actually intended.

PR #567, opened 2019-10-03
react native integration tutorial

mateosilguero:
## Summary

Integrate Flipper with React Native, is a bit different than a native app. I wrote this tutorial inspired by @nparashuram 's blog. http://blog.nparashuram.com/2019/09/using-flipper-with-react-native.html

## Changelog

Flipper + React Native Documentation

PR #569, opened 2019-10-03
Update CocoaPods in Getting Started documentation

zvona:
## Summary
"CocoaPods / Podfile" section in "Getting Started" documentation is out of sync with Podfile provided by React Native 0.61.x. This PR updates the necessary parts of setup. In addition, it does some clarifications in code.

## Changelog

Update section "CocoaPods" in "Getting Started" documentation.

## Test Plan

These are changes in documentation. The changes can be tested by installing pods and building the project.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -111,16 +111,23 @@ We support both Swift and Objective-C for Flipper with CocoaPods as build and di
 <!--Objective-C-->
 ```ruby
 project 'MyApp.xcodeproj'
-flipperkit_version = '0.25.0'
+
+def flipper_pods()
+  flipperkit_version = '0.25.0'
+  pod 'FlipperKit', '~>' + flipperkit_version, :configuration => 'Debug'
+  pod 'FlipperKit/FlipperKitLayoutComponentKitSupport', '~>' + flipperkit_version, :configuration => 'Debug'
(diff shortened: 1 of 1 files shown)


passy:
This looks great from what I can tell (which isn't much when it comes to CocoaPods, to be frank). Thanks for the contribution. I'll leave it to @priteshrnandgaonkar to review this. :)

priteshrnandgaonkar:
Although the current changes which you have made seems that it may work for the iOS apps. But it is worth checking out by changing the Sample App's podfile and building it, to verify if the setup works or not.

If it works then we can change the iOS integration docs along with the sample app. We plan to keep sample app as close to the documentation as possible.

zvona:
> Thanks for the PR. I would request you to add this documentation in the RN section instead of changing it the current iOS documentation.

so the idea would be having a separate section for RN config? If so, then this can be closed for sure and I'll create a new PR with a section with correct (and tested) config.

priteshrnandgaonkar:
> so the idea would be having a separate section for RN config? If so, then this can be closed for sure and I'll create a new PR with a section with correct (and tested) config.

Let's have a separate RN section with its installation instruction. It would be better to also have a sample app for RN which follows the installation instruction, so that users can directly tinker with it to understand the features of the flipper.

Closing this PR.

priteshrnandgaonkar:
Looks like #567 also added the integration instruction for iOS.

Issue #450, opened 2019-05-19
Discussion: [Plugin][Android] An other constructor for SharedPreferencesFlipperPlugin 


s1rius:
Hi, Can we create an other constructor for SharedPreferencesFlipperPlugin ?

Because I have a custom SharedPreferences implementation in a project. I cant just pass SharedPreferences's mode, name or a `SharedPreferencesDescriptor `.

Maybe we can add a constructor like below
```
  /**
   * Creates a {@link android.content.SharedPreferences} plugin for Flipper
   *
   * @param sharedPreferences The all SharedPreferences pair.
   */
  public SharedPreferencesFlipperPlugin (
      Map<SharedPreferences, SharedPreferencesDescriptor> sharedPreferences) {
    mSharedPreferences = sharedPreferences;
    ...
    registerOnSharedPreferenceChangeListener
    ...
  }

```
(And thanks build this tool)

bizzguy:
How would you see the UI for the plugin changing (if at all)?  Would this just be an app "filter" on the shared preference files you would see in the plugin?

passy:
I would think that should be compatible with what we currently have.

The `SharedPreferencesFlipperPlugin(Context context, List<SharedPreferencesDescriptor> descriptors)` effectively has the same result but you skip the resolution step. @s1rius if you want to put this up as a PR, I'm happy to take a look!

s1rius:
Hi @passy ,I rethink it and find, we can simple add a function to create SharedPreferences, and no need to add an other constructor in the SharedPreferencesFlipperPlugin Class.
So, I create a pull request, hope you take a look.

PR #571, opened 2019-10-03
Work with custom SharedPreferences implementation

s1rius:
## Summary
see #450 
If we want to control a custom SharedPreferences, we can create a Class as SharedPreferencesDescriptor's Subclass, override the getSharedPreferences() function to return a custom SharedPreferences instance.

## Changelog
add getSharedPreferences() function in SharedPreferencesDescriptor Class

## Test Plan
No


passy:
Yeah, that should work. Thanks for the contribution!

PR #574, opened 2019-10-08
added examples folder, with a RN example project

mateosilguero:
## Summary

Complementary with #567, i want to add an example project to show how to integrate Flipper and React Native.

## Changelog

Examples folder


mweststrate:
Sorry, totally missed this PR before! But we did create an example in our repo recently, which we will maintain and keep up to date, it will be compatible with 0.62 once released: https://github.com/facebook/flipper/tree/master/react-native/ReactNativeFlipperExample

PR #608, opened 2019-10-27
Corrected schedule spelling

umasankar-yedida:
Fixed a type in Logs Plugin


passy:
Good catch! Thank you.

Issue #326, opened 2018-11-14
Add a way to easily only see logs of the selected app


cristan:
What is shown in the logs screen by default is pretty much useless: there are so many logs that the logs of your app quickly scroll away.

This is easily fixed in Android Studio: select the package of your app, and select the Regex "Show only selected application". In Stetho, not so much. With a right click, you can select PID, find a log by your app and click on this PID.

This isn't very user friendly. Another problem is that the PID of your app isn't consistent, so after a reboot or when plugging in another device, you'll have to find the correct PID again.

Therefore: please add a way to easily see the logs of a specific app which has the Flipper SDK installed.

danielbuechele:
That's a really good point, which we'd like to improve. We need to find a good way to know which app the logs are coming from and also handle if an app is running multiple processes.

michaelcarrano:
Was just going to create a Github Issue about this. 

Would be great to filter the logs based on a selected application, even if I need to type the package name myself.

Not sure if it will help in anyway but Pidcat does this:
https://github.com/JakeWharton/pidcat

umasankar-yedida:
I'll take this up

nubpro:
What happened to this feature request?

vkotovv:
@nubpro Looks like it is stalled.

troy-lamerton:
Filtering by selected application is an essential feature.

Android Studio logcat:
![Screen Shot 2021-08-12 at 07 25 43](https://user-images.githubusercontent.com/13446008/129142717-2f7b4ee8-30d7-4913-b5ec-51ab8cb31c90.png)

Choosing application to filter by is a main feature of Android Studio's logcat.
I only care about logs from my application, not all this stuff from android system and other apps.

> We need to find a good way to know which app the logs are coming from and also handle if an app is running multiple processes.

1. `adb shell pidof my.app.package` -> prints a process id
2. filter the logs for that process id

@danielbuechele I think `pidof` will print only the main application process (not sure). If people want logs from all of an app's processes, that can be a new issue.

It is surprising that this issue has been stalled for so long. Without app filtering, I personally find that Flipper is too difficult to use.

PR #616, opened 2019-10-30
Enabled app column by fetching process name from pid in android

umasankar-yedida:
## Summary
This solves #326 

## Changelog
Fetching process name from pid when an entry is getting added to the table and adding it as a column called `app` (it already exists, but disabled for android).

Now, people can filter logs based on the `app` either by directly clicking on it from table or by typing `app=<process-name>` in search bar & hit enter

### DIFF
--- a/src/devices/AndroidDevice.tsx
+++ b/src/devices/AndroidDevice.tsx
@@ -166,7 +166,8 @@ export default class AndroidDevice extends BaseDevice {
   }
 
   async getProcessName(pid: number): Promise<string> {
-    return await this.adb.shell(this.serial, "ps | grep '" + pid + "'")
+    return await this.adb
+      .shell(this.serial, "ps | grep '" + pid + "'")
       .then(adbkit.util.readAll)
       .then(buffer => buffer.toString())
       .then(output => {
--- a/src/devices/AndroidDevice.tsx
+++ b/src/devices/AndroidDevice.tsx
@@ -167,15 +167,15 @@ export default class AndroidDevice extends BaseDevice {
 
   async getProcessName(pid: number): Promise<string> {
     return await this.adb.shell(this.serial, "ps | grep '" + pid + "'")
-        .then(adbkit.util.readAll)
-        .then(buffer => buffer.toString())
-        .then(output => {
-          const index = output.toString().lastIndexOf(" ");
-          if (index > -1) {
-            return output.substr(index + 1).trim();
-          } else {
-            return "unknown-process";
-          }
-        });
+      .then(adbkit.util.readAll)
+      .then(buffer => buffer.toString())
+      .then(output => {
+        const index = output.toString().lastIndexOf(' ');
+        if (index > -1) {
+          return output.substr(index + 1).trim();
+        } else {
+          return 'unknown-process';
+        }
+      });
   }
 }
--- a/src/devices/AndroidDevice.tsx
+++ b/src/devices/AndroidDevice.tsx
@@ -8,7 +8,8 @@
  */
 
 import BaseDevice, {DeviceType, LogLevel} from './BaseDevice';
-import adb, {Client as ADBClient} from 'adbkit';
+import adb from 'adbkit';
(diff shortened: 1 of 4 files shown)


src/devices/AndroidDevice.tsx
```diff
@@ -163,4 +164,19 @@ export default class AndroidDevice extends BaseDevice {
     this.recordingProcess = undefined;
     return destination;
   }
+
+  async getProcessName(pid: number): Promise<string> {
+    return await this.adb
```
priteshrnandgaonkar: I tried running Flipper with your changes, the performance of the app took a hit. It became unresponsive. I reckon, it is because of the `await` here. We wait for the `getProcessName` to finish and then we call `processEntry`. I will request you consider going for non blocking approach and avoid await.

umasankar-yedida: @priteshrnandgaonkar I have another approach in mind. But, I felt the way I have done is more intuitive for users. BTW, how did you monitor the performance? I will use the same once I do modifications

priteshrnandgaonkar: I didn't use any tool. I just ran Flipper and tried to use it normally, it was unresponsive.

jknoxville: The problem here is that this function is being called for every single log line, so when hundreds are coming in, it's starting up ps and grep processes many times per second. This isn't necessary - we don't need to recheck the pid again for every single log.

A better approach would be to memoize and reuse the process name for each device + pid, bearing in mind that they will occasionally change, so it should not cache forever.

In addition, if the adkbit promise takes a while to return, many log listener process entry calls are getting queued up until it resolves. I'm not sure if this is a problem in practice. I'd start by fixing the above one - I suspect thats the main issue.

src/plugins/logs/index.tsx
```diff
@@ -489,9 +489,17 @@ export default class LogTable extends FlipperDevicePlugin<
     };
 
     this.logListener = this.device.addLogListener((entry: DeviceLogEntry) => {
-      const processedEntry = processEntry(entry, String(this.counter++));
-      this.incrementCounterIfNeeded(processedEntry.entry);
-      this.scheduleEntryForBatch(processedEntry);
+      // Get the Process name from `pid` and add the process name as App column
+      const {pid} = entry;
+
+      this.device.getProcessName(pid).then(processName => {
```
jknoxville: I like the simplicity of the approach here, I hadn't thought of doing this, and it will be genuinely useful!

src/devices/BaseDevice.tsx
```diff
@@ -123,6 +123,10 @@ export default class BaseDevice {
     this._notifyLogListeners(entry);
   }
 
+  getProcessName(pid: number): Promise<string> {
+    return Promise.reject(() => new Error('Not supported'));
+  }
+
```
jknoxville: If I understand correctly, this will break the logs plugin for all device types other than android, e.g. iOS.

You could fix it by returning 'unknown-process' or similar.

umasankar-yedida: Got it. Will change it.

src/devices/AndroidDevice.tsx
umasankar-yedida: The reason why it is being called for every single log line is, at some point some process gets killed and the same `pid` can be assigned it to different process. So, the complexity here to memoize is, I can't make pid or process name as key. I tried to use `Select` component and once we select a process from that list I tried to filter the rows. Still, the application is becoming unresponsive after a minute or two. If we can get the process name from `adbkit` itself, I think we can gain better performance. @priteshrnandgaonkar @jknoxville Can you suggest?

troy-lamerton: Filtering by selected application is a killer feature. I would start using Flipper again once this is finished.

> If we can get the process name from adbkit itself, I think we can gain better performance.

@jknoxville what do you think? Is that a good approach?

Personally, I don't know this part of the project.

passy:
Closing as this has been stale for a long time.

Issue #470, opened 2019-06-18
Shared Preferences: Changelog updates aren't reflected in the inspector


mitchware-cfa:
## 🐛 Bug Report
When `SharedPreferences` updates are sent back to Flipper, the "current" values displayed in the inspector are not updated to reflect the changes in the changelog. 

## To Reproduce
- Setup Flipper with an Android sample app, a `SharedPreferences` reference, and the `SharedPreferences` flipper plugin installed.
- Set a value in the preferences
- Update that value at runtime and observe the changelog in flipper

## Expected Behavior
I expect the inspector values to update in real time in response to the changes sent back from the mobile plugin

## Environment
- Android OS version 9.0
- Flipper SDK version 0.15.0
- Flipper Desktop version 0.21.1
- Android gradle plugin version 3.4.1

umasankar-yedida:
I'm working on it

vkotovv:
Is this still not resolved?

passy:
Thanks for noticing. That was addressed with #623.

PR #623, opened 2019-11-03
Producing new state when any change to preferences happen to address #470

umasankar-yedida:
## Summary

Fixed issue #470 

## Changelog

Producing new state when preferences gets updated in SharedPreferences plugin

### DIFF
--- a/src/plugins/shared_preferences/index.js
+++ b/src/plugins/shared_preferences/index.js
@@ -20,7 +20,7 @@ import {
 } from 'flipper';
 import {FlipperPlugin} from 'flipper';
 
-const {clone} = require('lodash');
+import {clone} from 'lodash';
 
 type SharedPreferencesChangeEvent = {|
   preferences: string,
@@ -97,8 +97,12 @@ export default class extends FlipperPlugin<SharedPreferencesState> {
       entry.preferences = update.preferences;
       state.sharedPreferences[update.name] = entry;
       return {
+        ...state,
         selectedPreferences: state.selectedPreferences || update.name,
-        sharedPreferences: state.sharedPreferences,
+        sharedPreferences: {
+          ...state.sharedPreferences,
+          [update.name]: entry,
+        },
       };
     },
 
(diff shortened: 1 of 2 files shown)


jknoxville:
Thanks for the contribution @umasankar-yedida 
Perhaps we should for pragmatism but I'm a little bit hesitant to merge as this seems to be hiding the issue of the DataInspector not getting re-rendered when it should be.
However, I've spent some time looiking into it and I'm still not sure why that is. @umasankar-yedida or @mweststrate do you know what's going on here?

mweststrate:
The problem here is not in the DataInspector (this change would kill most optimizations for all consumers of that component!). The problem is in the reducers of the shared preferences plugin, which does not produce new state, but rather modifies it's current state: 

https://github.com/facebook/flipper/blob/6e8de457e7882ad51de1e44ad9fa9d82ece2fa75/src/plugins/shared_preferences/index.js#L94-L121

I'd expect making the following change fixes the problem (sorry, I am not sure yet how to test that change :)) 

```
    UpdateSharedPreferences(state: SharedPreferencesState, results: Object) {
      const update = results.update;
      const entry = state.sharedPreferences[update.name] || {changesList: []};
      entry.preferences = update.preferences;
      // don't mutate the current state!
      return {
         ...state,
         selectedPreferences: state.selectedPreferences || update.name,
         sharedPreferences: {
             ...state.sharedPreferences,
            [update.name] = entry;
         }
      }
    },

    ChangeSharedPreferences(state: SharedPreferencesState, event: Object) {
      const change = event.change;
      const entry = state.sharedPreferences[change.preferences];
      if (entry == null) {
        return state;
      }
      let newEntry
      if (change.deleted) {
        const newPreferences = {...entry.preferences }
        delete newPreferences[change.name]
        newEntry = { ...entry, preferences: newPreferences }
      } else {
        newEntry = { ...entry, preferences: { ...entry.preferences, [change.name] : change.value } };
      }
      newEntry.changesList = [change, ...entry.changesList];
      return {
        ...state,
        sharedPreferences: {
            ...state.sharedPreferences,
            [change.preferences] = newEntry
        }
      };
    },
```

umasankar-yedida:
From the issue, I understood that the props are not getting updated in `ManagedDataInspector` component as this is acting as a wrapper to the `DataInspector` component.

umasankar-yedida:
@mweststrate I tried your approach and it worked !!! Thanks man. :)

PR #715, opened 2019-12-20
fix(flipper-static): show default warning message on firing in unknown flags

jamesgeorge007:
## Summary

Passing in unknown options along with `flutter-static` now shows the default warning message as part of the API.

## Changelog
show default warning message on firing in unknown flags

## Test Plan
N/A

### DIFF
--- a/static/index.js
+++ b/static/index.js
@@ -70,6 +70,7 @@ const argv = yargs
   })
   .version(global.__VERSION__)
   .help()
+  .strict()
   .parse(process.argv.slice(1));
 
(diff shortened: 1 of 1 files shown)


passy:
Thanks for the submission, but we had to revert this because this conflicts with a bunch of internal processes that we have.

jamesgeorge007:
@passy that's odd. Can you lemme know what went wrong?

passy:
`yarn start` stopped working and in addition to this we have some scripts which must continue to work in forwards and backwards-compatible ways where flags may be added and removed.

PR #738, opened 2020-01-09
persist favorites queries at localStorage. show parsed values from databases.

mateosilguero:
## Summary

on databases section, queries marked as "favorite", are not persisted. With this changes,  they will be saved at localStorage and persist from refresh.

Improves on parse json values from databases.


src/plugins/databases/index.js
```diff
@@ -783,6 +784,10 @@ export default class DatabasesPlugin extends FlipperPlugin<
             newFavorites = state.favorites.concat(value);
           }
         }
+        window.localStorage.setItem(
+          'favoritesQueries',
```
passy: It would be nice to have something a little more specific as `favoriteQueries` could apply to a variety of plugins we have. I don't know off the top of my head if we've got any better places for storing information that's supposed to persist across sessions, but I'll ask internally.

src/plugins/databases/index.js
```diff
@@ -1071,13 +1080,32 @@ export default class DatabasesPlugin extends FlipperPlugin<
...
+        for (const key in parsed) {
+          try {
+            sidebarArray.push(
+              this.buildSidebarRow(key, {
+                props: {
+                  children: parsed[key],
+                },
+              }),
+            );
+          } catch (err) {}
```
passy: What's the error you expect here?

src/plugins/databases/index.js
```diff
@@ -940,6 +945,10 @@ export default class DatabasesPlugin extends FlipperPlugin<
         databases,
       });
     });
+    this.dispatchAction({
+      type: 'UpdateFavorites',
+      favorites: JSON.parse(localStorage.getItem('favoritesSQLQueries')),
```
passy: There are a couple of flow errors which somehow aren't shown in CI (will investigate that separately).

This may fail because `localStorage.getItem` can return undefined. I'd just turn this into `... || '[]'`.

passy:
re: the missing Flow checks. If you rebase this on top of master (or merge it in), you should be getting Flow checks as part of the CI signals here.

mateosilguero:
CI crash because i commit some unused files without intention 😅 now its all green

PR #740, opened 2020-01-13
iOS/Sample : UI need to be updated on Main Thread

0xd-cc:
### iOSSample:
Fix iOS Sample App crash: UI need to be updated on Main Thread

### SampleSwift:
* Upgrade swift version
* Fix force unwrapping
* UI need to be updated on Main Thread


passy:
Thanks for the PR! I'm definitely not the best person to review this, but I'll import that and find someone internal to take a look.

PR #760, opened 2020-01-25
Make database plugin query textarea resizable

angusholder:
## Summary

In the database plugin, make the SQL text entry box vertically-resizable to improve UX when working with longer queries.

I had wanted to make it a horizontal drag-to-resize divider like in DB Browser for SQLite (see below) as that feels like more what I'd expect to find for this use-case, but it seems to be [non-trivial in HTML and CSS](https://stackoverflow.com/questions/38983039/how-to-create-vertical-drag-to-resize-divider).

image

Side-note, is there any good existing component (in React or Flipper itself) for doing query tabs like in the above screenshot?

## Changelog

Make the database SQL text box vertically-resizable

## Test Plan

I wrote out a long query and was able to resize the text box to view the whole thing at once
image

### DIFF
--- a/src/plugins/databases/index.js
+++ b/src/plugins/databases/index.js
@@ -1302,6 +1302,7 @@ export default class DatabasesPlugin extends FlipperPlugin<
                   marginLeft: 16,
                   marginTop: '1%',
                   marginBottom: '1%',
+                  resize: 'vertical',
                 }}
                 onChange={this.onQueryChanged.bind(this)}
                 placeholder="Type query here.."


mweststrate:
Thanks for the PR! Landing this now. 

For your question: Yes, there is `TabsContainer` and `Tabs`: https://fbflipper.com/docs/extending/ui-components.html#tabscontainer

jknoxville:
> Side-note, is there any good existing component (in React or Flipper itself) for doing query tabs like in the above screenshot?

The nearest thing we have are TabsContainer and the [Tabs](https://fbflipper.com/docs/extending/ui-components.html#tabs) component.

You can see them in action by clicking "Manage Plugins" in flipper, and look at the code here: https://github.com/facebook/flipper/blob/master/src/chrome/PluginManager.tsx#L30

PR #761, opened 2020-01-25
Enable using Ctrl+Return to execute a query in the database plugin

angusholder:
## Summary

Having a key shortcut to execute the query you're typing helps to speed up the edit-run cycle.

## Changelog

In the database SQL execution tab, you can now execute queries with Ctrl+Return instead of pressing the 'Execute' button.

## Test Plan

I typed a query into the textarea, and while it was focussed, press Ctrl+Return, and the query executed. 

Mousing over the 'Execute' button displays a tooltip informing you of this keyboard shortcut.

image


passy:
Looks good to me! Thanks for the UX improvement!

PR #762, opened 2020-01-27
Remove stetho

### LINKED ISSUE #607: Discussion: Consider moving stetho support to a separate artifact
Stetho was never updated to androidx. This means that flipper and any consumers of it have to use jetifier still because of it _even if_ they don't use stetho and _even though_ flipper itself has moved to androidx.

```
* com.facebook.flipper:flipper:0.26.0
  \-- com.facebook.stetho:stetho:1.5.1
   \-- com.android.support:support-annotations:28.0.0

* com.facebook.flipper:flipper-network-plugin:0.26.0
  \-- com.facebook.flipper:flipper:0.26.0
   \-- com.facebook.stetho:stetho:1.5.1
    \-- com.android.support:support-annotations:28.0.0
```

ZacSweers:
## Summary 

Resolves #607 

This removes stetho dependencies directly in favor of just copying in the used classes. This removes the last non-androidx dependency from the project, and is useful as stetho appears to be unmaintained. The fresco stetho dependency appeared to be completely unused.

## Changelog

Remove stetho dependencies. It was unused other than a few utility methods, but came at the cost of preventing Flipper consumers from completely removing dependencies that hadn't updated to Androidx yet.

## Test Plan

Existing tests still pass with no regressions. Functionally this should be no difference.


passy:
Brilliant! Thanks for doing this. I may not have time to review this today, but will take a look soon. Greatly appreciated. :)

ZacSweers:
Sounds good. It looks like the fresco integration was using stetho, I can take a look at patching things up

android/plugins/fresco/build.gradle
```diff
@@ -21,7 +21,6 @@ android {
         implementation project(':android')
         implementation deps.fresco
         implementation deps.frescoFlipper
-        implementation deps.frescoStetho
```
passy: This doesn't seem to be quite covered: https://travis-ci.org/facebook/flipper/jobs/642213519?utm_medium=notification&utm_source=github_status

I wonder if this is just a transitive dependency and not actually a dependency on the Stetho parts in it: https://repo1.maven.org/maven2/com/facebook/fresco/stetho/2.0.0/stetho-2.0.0.pom

passy: Had a chat with @oprisnik from Fresco. Currently, the Cache class causing the failure is part of the Stetho artefact, but for no other reason than convenience when we first did this. He'll try to find somebody to create a separate package for this.

In the meantime, @ZacSweers, would you mind trying to use an exclude rule for on this dependency on this? I haven't tried it but excluding the Stetho dep should still work, e.g.

```groovy
implementation(deps.frescoStetho) {
   exclude group: 'com.facebook.stetho'
}
```

(From memory, might be the wrong syntax.)

ZacSweers: Updated 👍  d4390d3

oprisnik: Thanks. That should work for now. We'll do a refactor on our side soon so that our Flipper plugin doesn't depend on Stetho. Once we've released a new Fresco version with that fix, this can be simplified again :)

oprisnik: The Stetho dep has been removed: https://github.com/facebook/fresco/commit/d1d5bdaf8c37058bf965920d7caa461a8402ef98

The next Fresco release will include this change.

PR #765, opened 2020-01-28
Flipper: Fix exception in popup menus

rgiese:
Summary: When JavaScript fiction meets native interop reality, things get weird. Apparently some coordinates must be integer values.

Reviewed By: mweststrate

Differential Revision: D19606677


passy:
This was merged but somehow not closed here.

PR #795, opened 2020-02-12
Implement react-native-flipper on iOS

janicduplessis:
## Summary

Implement the react-native-flipper native module on iOS. Uses very similar abstractions as on Android.

## Changelog

[react-native-flipper] Support iOS

## Test Plan

Tested using the RN TicTacToe example app


janicduplessis:
Some cleaning up to do still, also need to figure out what to do with `reportError*` methods that don't have an equivalent on the iOS FlipperConnection class.

passy:
Wow! 🎉 I'm not the right person to review iOS stuff, but this looks amazing and was actually something that we had on our roadmap. Thanks a lot for working on this, @janicduplessis!

react-native/ReactNativeFlipperExample/ios/Podfile
```diff
@@ -70,12 +70,8 @@ target 'ReactNativeFlipperExample' do
...
   pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec'
 
-  target 'ReactNativeFlipperExampleTests' do
-    inherit! :complete
-    # Pods for testing
-  end
-
-  use_native_modules!
+  # use_native_modules!
+  pod 'react-native-flipper', :path => "../../react-native-flipper", :configuration => 'Debug'
```
thymikee: How about not disabling autolinking and use configure CLI to see local lib? https://github.com/react-native-community/cli/blob/master/docs/autolinking.md#how-can-i-autolink-a-local-library

mweststrate: note that relative-deps has been set up in this example project, so running `yarn relative-deps` in the root should already copy the module to make it local (assuming that works for iOS as well, it works fine for Android)

thymikee: Hm, what I mean is that we don't need to disable autolinking (commenting out the `use_native_modules!`). With the right config, we could also get rid of the `pod` declaration, but that's fine and actually less code :D

priteshrnandgaonkar:
I imported the PR and it fails to build the ReactNativeFlipperExample app, seems like the react-native-flipper pod doesn't have FlipperKit as a dependency. I will do some changes on your PR and make it work.

janicduplessis:
@priteshrnandgaonkar I also had some issues with the example app setup. I think I ended up modifying the generated pod target for react-native-flipper to add the FB_SONAR_ENABLED flag manually because it wasn’t there for some reason.

react-native/ReactNativeFlipperExample/ios/Podfile
janicduplessis: Yea the example app setup changes were temporary just so I can get it building. The problem I had is I wanted to open the xcodeworkspace for the example app and edit files from there but using autolinking + relativedeps it would use a copy under node_modules instead. Or at least thats what I thought it did.

priteshrnandgaonkar:
You had to set ` FB_SONAR_ENABLED=1` in the compiler flag of your pod. Also I have added the implementation of `reportError` and `reportErrorWithMetadata`. Also I feel the name of the pod should be in camel case as that is what the convention is in iOS. Since its a sample app, it makes sense to locally refer FlipperKit pod rather than referring the prod one. I will do those changes too.

janicduplessis:
@priteshrnandgaonkar Awesome! Feel free to finish up what is missing and land it.

janicduplessis:
@priteshrnandgaonkar Awesome! Feel free to finish up what is missing and land it.

PR #819, opened 2020-02-21
[flipper] Kill orphaned instruments processes

### LINKED ISSUE #808: Flipper spawns hundreds of DTServiceHub processes over time
## 🐛 Bug Report
Over time Flipper spawns more and more "DTServiceHub" processes. To the point where it bricks my system

image
image

## To Reproduce
Seems to happen when Flipper is running for a long time. I can't really find a step-by-step reproduction. 

Note: I only had the 'network' plugin enabled in the desktop app when this occured. But I can also see it spawning processes when no plugins are enabled at all.

## Expected Behavior
To not brick my macbook by spawning so many processes that MacOS can't spawn any new ones.

## Environment
MacOS Catalina `10.15.3 (19D76)`
Flipper desktop app `Version 0.31.1 (0.31.1)`
Simulator `Version 11.3.1 (SimulatorApp-912.5.1 SimulatorKit-570.3 CoreSimulator-681.17.2)`
Xcode `Version 11.3.1 (11C505)`

`flipperkit_version = '0.31.1'`

jknoxville:
Fix #808 
Kills any orphaned Instruments processes belonging to the user.

In some cases, we've seen interactions between Instruments and the iOS
simulator that cause hung instruments and DTServiceHub processes. If
enough instances pile up, the host machine eventually becomes
unresponsive. Until the underlying issue is resolved, manually kill any
orphaned instances (where the parent process has died and PPID is 1)
before launching another instruments run.

Taking the same approach as done by flutter here.

Facebook: This is doing in OSS, what was already done internally. Will follow up with a diff that shares the code.

### DIFF
--- a/src/fb-stubs/iOSContainerUtility.tsx
+++ b/src/fb-stubs/iOSContainerUtility.tsx
@@ -10,6 +10,7 @@
 import {DeviceType} from '../devices/BaseDevice';
 import {exec} from 'promisify-child-process';
 import {notNull} from '../utils/typeUtils';
+import {killOrphanedInstrumentsProcesses} from '../utils/processCleanup';
 
 const errorMessage = 'Physical iOS devices not yet supported';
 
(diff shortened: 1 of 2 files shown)


passy:
Got landed.

PR #822, opened 2020-02-23
Add React Native/Metro hotkeys

### LINKED ISSUE #798: Discussion: add hotkeys to trigger Metro commands
It would be nice to add hotkeys to trigger the functionalities added in [`3dcfe9f`](https://github.com/facebook/flipper/commit/3dcfe9f3aef2b02ec8f520c0d331a962c1a1e197) for reloading and opening the developer menu.

It would also be great if those hotkeys were global and could, perhaps, be configured in the preferences screen, I would definitely be in for working on that.

lucasbento:
## Summary

This PR fixes #798 by adding customizable hotkeys to reload and/or open developer menu in React Native apps.

Screenshot of the Preferences window with hotkeys

#### TODO:

- [x] Add correct icon for removing content of the hotkey input (currently using `undo`) - cc @passy 😄 

## Changelog

Add customizable hotkeys to reload and/or open developer menu in React Native apps.

## Test Plan

- Run React Native on version `0.62.0-rc.2` (you can use this app: https://github.com/lucasbento/RNWithFlipper);
- Open the Preferences window (`⌘,`);
- Customise the React Native hotkeys to whatever you want;
- Test them out with Flipper's window active and inactive.

> **Note**: this has been tested only in macOS.

### DIFF
--- a/src/chrome/settings/KeyboardShortcutInput.tsx
+++ b/src/chrome/settings/KeyboardShortcutInput.tsx
@@ -242,7 +242,7 @@ const KeyboardShortcutInput = (props: {
             character: '',
           })
         }>
-        <CenteredGlyph name="undo" variant="outline" />
+        <CenteredGlyph name="cross" variant="outline" />
       </FlexColumn>
     </Container>
   );
--- a/src/chrome/settings/KeyboardShortcutInput.tsx
+++ b/src/chrome/settings/KeyboardShortcutInput.tsx
@@ -1,0 +1,251 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ */
+
+import {FlexColumn, styled, FlexRow, Text, Glyph, colors} from 'flipper';
+import React, {useRef, useState, useEffect} from 'react';
+
+type PressedKeys = {
+  metaKey: boolean;
+  altKey: boolean;
+  ctrlKey: boolean;
+  shiftKey: boolean;
+  character: string;
+};
+
+const KEYCODES = {
+  DELETE: 8,
+  ALT: 18,
+  SHIFT: 16,
+  CTRL: 17,
+  LEFT_COMMAND: 91, // Left ⌘ / Windows Key / Chromebook Search key
+  RIGHT_COMMAND: 93, // Right ⌘ / Windows Menu
+};
+
+const ACCELERATORS = {
+  COMMAND: 'Command',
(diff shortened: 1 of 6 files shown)


src/chrome/settings/KeyboardShortcutInput.tsx
```diff
@@ -0,0 +1,251 @@
...
+        onClick={() =>
+          handleUpdatePressedKeys({
+            metaKey: false,
+            altKey: false,
+            ctrlKey: false,
+            shiftKey: false,
+            character: '',
+          })
+        }>
+        <CenteredGlyph name="undo" variant="outline" />
```
passy: Screenshot 2020-02-23 at 16 29 45

Do you think one of these could work?

lucasbento:
> One thing I wonder is if we want to make this opt-in or opt-out. As they're global I tend towards the former, but I'm curious to hear your thoughts on this.

I don't fully get it, do you mean that it should be turned off by default? 

I see that there was an error with the actions: https://github.com/facebook/flipper/pull/822/checks?check_run_id=463154571 do you know what the problem is?

src/chrome/settings/KeyboardShortcutInput.tsx
lucasbento: Yeah, the `cross` will mostly definitely work, thank you!

I'll change it today in the evening.

passy:
@lucasbento 

> I don't fully get it, do you mean that it should be turned off by default?

Sorry, I tend to not make sense on Sundays. 😅 I think off by default may be preferable because RN is already only a subset of the users and there's a chance - even slim - that we might interfere with other shortcuts people could have. We could think about advertising them when using Metro devices, though, which would be great even if enabled by default to make people aware of them. (Definitely as a separate task, though.)

> I see that there was an error with the actions:

There's a fix waiting to be landed. We've had some infra issues late last week that meant that all internal changes got stuck, but that'll land in just a bit. :)

lucasbento:
@passy:

> I think off by default may be preferable because RN is already only a subset of the users and there's a chance - even slim - that we might interfere with other shortcuts people could have. We could think about advertising them when using Metro devices, though, which would be great even if enabled by default to make people aware of them. (Definitely as a separate task, though.)

Makes sense, would you mind creating an issue for this just so I don't forget? I'm gonna work on it after this is landed.

> There's a fix waiting to be landed. We've had some infra issues late last week that meant that all internal changes got stuck, but that'll land in just a bit. :)

Good to know!

lucasbento:
@passy: all done :)

It's also disabled by default as we agreed.

passy:
You're a legend! Thanks for making the changes!

PR #874, opened 2020-03-06
Make iOS 9 compatible again.

alloy:
## Summary

React Native v0.62.0’s template still uses iOS 9 as deployment target, bumping it to 10 this late in the release cycle is probably not a good idea, so instead I made this tiny change to make it iOS 9 compatible again.

⚠️  I made this change on top of the `v0.32.2` **tag**, as `master` was giving me a build failure and in any case it would be better to release _just_ this change as a patch release so the RN `v0.62.0-rc.3` release can go out without having to test Flipper as thoroughly again.

## Changelog

Make FlipperKit compatible with iOS 9 again.

## Test Plan

- `pod install` with a Podfile that has iOS 9 as its deployment target passes again.
- Building the Flipper Sample works.
- Building with a RN application created with the `v0.62.0-rc.3` template works.

### DIFF
--- a/iOS/Plugins/FlipperKitLayoutPlugin/FlipperKitLayoutPlugin/descriptors/SKViewDescriptor.mm
+++ b/iOS/Plugins/FlipperKitLayoutPlugin/FlipperKitLayoutPlugin/descriptors/SKViewDescriptor.mm
@@ -222,14 +222,12 @@ static NSDictionary* YGUnitEnumMap = nil;
           nil];
 }
 
-- (NSDictionary<NSString*, SKNodeUpdateData>*)dataMutationsForNode:
-    (UIView*)node {
-  return @{
+- (NSDictionary<NSString*, SKNodeUpdateData>*)dataMutationsForNode:(UIView*)node {
+  NSDictionary<NSString*, SKNodeUpdateData> *dataMutations = @{
     // UIView
     @"UIView.alpha" : ^(NSNumber* value){
         node.alpha = [value floatValue];
-}
-,
+    },
     @"UIView.backgroundColor": ^(NSNumber *value) {
       node.backgroundColor = [UIColor fromSonarValue: value];
     },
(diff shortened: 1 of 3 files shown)


alloy:
Note that even with this change, there is API being used in Folly that’s also iOS 10 only, but that was also the case with FlipperKit `v0.30.2`, so perhaps this code-path is simply never hit from FlipperKit?

Screenshot 2020-03-06 at 17 06 14

alloy:
I see that the cause of the build failure on master that I mentioned has already been fixed https://github.com/facebook/flipper/commit/987822074b87417c60f0248a1c19735e7cf0c730

iOS/Plugins/FlipperKitLayoutPlugin/FlipperKitLayoutPlugin/descriptors/SKViewDescriptor.mm
```diff
@@ -452,17 +450,21 @@ - (NSArray*)validChildrenForNode:(UIView*)node {
...
     },
-}
-;
+  };
+  if (@available(iOS 10.0, *)) {
+    NSMutableDictionary<NSString*, SKNodeUpdateData> *latestDataMutations = [dataMutations mutableCopy];
+    latestDataMutations[@"Accessibility.accessibilityTraits.UIAccessibilityTraitTabBar"] = ^(NSNumber *value) {
+      node.accessibilityTraits = AccessibilityTraitsToggle(node.accessibilityTraits, UIAccessibilityTraitTabBar, [value boolValue]);
+    };
+    dataMutations = [latestDataMutations copy];
```
passy: Happy to merge it like this but I got the feedback internally, that it would be fine to skip this as we have no further references to the mutable copy and could cast it.

alloy: Works for me; I’ll make the change 👍

alloy: Done ✅

PR #926, opened 2020-03-25
Add gitattributes with EOF settings

passy:
Summary:
This is to mitigate prettier problems on Windows as described here:
https://github.com/prettier/prettier/issues/7825#issuecomment-602171740

Test Plan:
Anton.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


nikoant:
@passy need to rebase this PR onto "master". Windows jobs are failing now because of another issue which is already fixed in "master'.

PR #941, opened 2020-03-29
Imported <DescriptorMapping> and <InspectorFlipperPlugin>

jeanniton-mnr:
Fixed cannot find symbols: `DescriptorMapping` and `InspectorFlipperPlugin`


passy:
Thanks for helping out with the docs! :)

Issue #931, opened 2020-03-26
Could not retrieve mAdded field from class androidx.fragment.app.FragmentManagerImpl


benoitdion:
## 🐛 Bug Report

It looks like we're looking for a `mAdded` field https://github.com/facebook/flipper/blob/48d6ea4c30c7593b7dace7335453e5b45baa4d2c/android/src/main/java/com/facebook/flipper/plugins/inspector/descriptors/utils/stethocopies/FragmentCompat.java#L87 on `FragmentManagerImpl` but that field doesn't exist: 
https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/fragment/fragment/src/main/java/androidx/fragment/app/FragmentManager.java
https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/fragment/fragment/src/main/java/androidx/fragment/app/FragmentManagerImpl.java
`mAdded` exists on the fragment themselves but not the manager. 

Would https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/fragment/fragment/src/main/java/androidx/fragment/app/FragmentManager.java#934 be useful for what we're trying to do here?

Stacktrace:
```
2020-03-26 13:01:23.808 8969-8969/com.mypackage E/com.facebook.flipper.plugins.inspector.descriptors.utils.stethocopies.ReflectionUtil: Could not retrieve mAdded field from class androidx.fragment.app.FragmentManagerImpl
    java.lang.NoSuchFieldException: No field mAdded in class Landroidx/fragment/app/FragmentManagerImpl; (declaration of 'androidx.fragment.app.FragmentManagerImpl' appears in /data/app/com.mypackage-5E855k89Sn36xw35D2A==/base.apk)
        at java.lang.Class.getDeclaredField(Native Method)
        at com.facebook.flipper.plugins.inspector.descriptors.utils.stethocopies.ReflectionUtil.tryGetDeclaredField(ReflectionUtil.java:29)
        at com.facebook.flipper.plugins.inspector.descriptors.utils.stethocopies.FragmentCompat$FragmentManagerAccessorViaReflection.getAddedFragments(FragmentCompat.java:88)
        at com.facebook.flipper.plugins.inspector.descriptors.utils.stethocopies.FragmentCompatUtil.findFragmentForViewInFragmentManager(FragmentCompatUtil.java:82)
[...]

PR #957, opened 2020-04-01
Add support for AndroidX fragments

passy:
Summary:
Fix https://github.com/facebook/flipper/issues/931

This is not how I would *like* to fix this, but it should do the job.
When the switch over to AndroidX was made, the overall abstraction
started to leak and we really need to remodel this in its entirety.
There's also the question of whether we want to support both support
fragments and AndroidX fragments or not. Right now it's kinda-sorta
supported but only under some circumstances, which is not great.

I also added some more defensive try/catches as there's some unsafe casting
involved and future changes may break this causing the entire layout to disappear.

Test Plan:
Changed the sample app to include some AndroidX fragments and they
now show up (again) in the view hierarchy:

Screenshot 2020-04-01 13 40 53

Change Log:
Fix support for AndroidX fragments in Layout Inspector.

### DIFF
--- a/android/src/main/java/com/facebook/flipper/plugins/inspector/descriptors/utils/stethocopies/FragmentCompat.java
+++ b/android/src/main/java/com/facebook/flipper/plugins/inspector/descriptors/utils/stethocopies/FragmentCompat.java
@@ -8,8 +8,14 @@
 package com.facebook.flipper.plugins.inspector.descriptors.utils.stethocopies;
 
 import android.app.Activity;
+import android.content.Context;
 import android.os.Build;
+import android.util.Log;
+
+import androidx.fragment.app.FragmentManager;
+
 import java.lang.reflect.Field;
+import java.util.Collections;
 import java.util.List;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/facebook/flipper/plugins/inspector/descriptors/utils/stethocopies/FragmentCompat.java
```diff
@@ -82,20 +88,22 @@ public static FragmentCompat getSupportLibInstance() {
...
     public List<FRAGMENT> getAddedFragments(FRAGMENT_MANAGER fragmentManager) {
-      // This field is actually sitting on FragmentManagerImpl, which derives from FragmentManager.
-      if (mFieldMAdded == null) {
-        Field fieldMAdded =
-            ReflectionUtil.tryGetDeclaredField(fragmentManager.getClass(), "mAdded");
-
-        if (fieldMAdded != null) {
-          fieldMAdded.setAccessible(true);
-          mFieldMAdded = fieldMAdded;
+      if (fragmentManager instanceof android.app.FragmentManager) {
```
benoitdion, line 91: The `instanceof` check here defeats the purpose of this abstraction a bit but this looks like a working fix otherwise.

benoitdion:
@passy will you also be committing the changes to the example app to help catch regressions sooner?

PR #978, opened 2020-04-05
Fix IOS crash in Network Plugin due to incorrect processing of data U…

### LINKED ISSUE #974: iOS - Network plugin - Crash when response is not an instance of NSHTTPURLResponse
## 🐛 Bug Report

I encountered the following crash: `Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[NSURLResponse allHeaderFields]: unrecognized selector sent to instance 0x600000d403a0'`

This crash is occurring in `SKResponseInfo.m` inside the `shouldStripReponseBodyWithResponse` method. It seems that the response is not an instance of `NSHTTPURLResponse` and that is the reason that `allHeaderFields` does not exist.

This seems to happen because the to be monitored request is an inline image data url (e.g, `data:image/x-png;base64,/...`). 

## To Reproduce
I'm using Flipper with React Native 0.62.  The crash occurs in debug builds when:
- From JS, 
-- Add an <Image /> with a an inline image data url
-- Invoke `Image.getSize(url)` where url is the given inline image data url.
- From Objective-C, 
-- `RCTDataRequestHandler` creates a network request
-- The request is intercepted by the FLEXNetworkRecorder from where the crash is triggered inside `SKResponseInfo`

As a temp workaround, I added checks if `[response isKindOfClass:[NSHTTPURLResponse class]]` is true, in `shouldStripReponseBodyWithResponse` and `didObserveResponse`.

## Environment
Flipper: 0.33.1
System:
    OS: macOS 10.15.3
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 142.72 MB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.14.0 - ~/.nvm/versions/node/v12.14.0/bin/node
    Yarn: 1.22.4 - ~/.yarn/bin/yarn
    npm: 6.13.6 - ~/.nvm/versions/node/v12.14.0/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.4, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK:
      API Levels: 23, 28, 29
      Build Tools: 28.0.3, 29.0.0, 29.0.1, 29.0.2
      System Images: android-28 | Google Play Intel x86 Atom, android-28 | Google Play Intel x86 Atom_64, android-29 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.6 AI-192.7142.36.36.6241897
    Xcode: 11.4/11E146 - /usr/bin/xcodebuild

mark-plukkido:
Fix IOS crash in Network Plugin due to incorrect processing of data URLs (#974)

## Summary
Fix #974 by skipping response processing in FlipperKitNetworkPlugin.mm if the response is not an instance of NSHTTPURLResponse, which data URLs are not. My assumption is that data URLs are not the ones Flipper Network Plugin users are interested in, given the type of information being extracted in the didObserveResponse method which are only present in NSHTTPURLReponse types.

## Changelog
Fix IOS crash in Network Plugin due to unchecked casting of data URLs

## Test plan
1. `npx react-native init issue974 --version react-native@0.62.1`
2. `cd issue974/ios/Pods/FlipperKit/iOS/Plugins/`
3. `curl -L https://patch-diff.githubusercontent.com/raw/facebook/flipper/pull/978.patch | git apply -v -p3`
4.  `cd ../../../../../`
5. `curl -L https://github.com/facebook/flipper/files/4434063/rn-data-uri-test.patch.txt | git apply -v
react-native run-ios`
6. Verify that the app does not crash the IOS app with 
```
Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 
'-[NSURLResponse allHeaderFields]: unrecognized selector sent to instance <memaddr>'
```

#### The React Native app patch used to verify the PR working and resolving the issue #974
[rn-data-uri-test.patch.txt](https://github.com/facebook/flipper/files/4434063/rn-data-uri-test.patch.txt)

### DIFF
--- a/iOS/Plugins/FlipperKitNetworkPlugin/FlipperKitNetworkPlugin/SKResponseInfo.m
+++ b/iOS/Plugins/FlipperKitNetworkPlugin/FlipperKitNetworkPlugin/SKResponseInfo.m
@@ -29,6 +29,11 @@
 }
 
 + (BOOL)shouldStripReponseBodyWithResponse:(NSURLResponse*)response {
+  // Only HTTP(S) responses have Content-Type headers
+  if (![response isKindOfClass:[NSHTTPURLResponse class]]) {
+    return NO;
+  }
+
   NSHTTPURLResponse* httpResponse = (NSHTTPURLResponse*)response;
   NSString* contentType = httpResponse.allHeaderFields[@"content-type"];
   if (!contentType) {
--- a/iOS/Plugins/FlipperKitNetworkPlugin/FlipperKitNetworkPlugin/SKResponseInfo.m
+++ b/iOS/Plugins/FlipperKitNetworkPlugin/FlipperKitNetworkPlugin/SKResponseInfo.m
@@ -31,7 +31,7 @@
 + (BOOL)shouldStripReponseBodyWithResponse:(NSURLResponse*)response {
   // Only HTTP(S) responses have Content-Type headers
   if (![response isKindOfClass:[NSHTTPURLResponse class]]) {
-    return NO;
+    return YES;
   }
 
   NSHTTPURLResponse* httpResponse = (NSHTTPURLResponse*)response;
--- a/iOS/Plugins/FlipperKitNetworkPlugin/FlipperKitNetworkPlugin/FlipperKitNetworkPlugin.mm
+++ b/iOS/Plugins/FlipperKitNetworkPlugin/FlipperKitNetworkPlugin/FlipperKitNetworkPlugin.mm
@@ -78,6 +78,11 @@
 }
 
 - (void)didObserveResponse:(SKResponseInfo*)response {
+  // Only track HTTP(S) calls, data URLs cannot be casted to NSHTTPURLResponse
+  if (![response.response isKindOfClass:[NSHTTPURLResponse class]]) {
+    return;
+  }
+
   NSHTTPURLResponse* httpResponse = (NSHTTPURLResponse*)response.response;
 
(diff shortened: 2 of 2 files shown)


cekkaewnumchai:
Thank you for the PR!

iOS/Plugins/FlipperKitNetworkPlugin/FlipperKitNetworkPlugin/SKResponseInfo.m
```diff
@@ -29,6 +29,11 @@ - (instancetype)initWithIndentifier:(int64_t)identifier
 }
 
 + (BOOL)shouldStripReponseBodyWithResponse:(NSURLResponse*)response {
+  // Only HTTP(S) responses have Content-Type headers
+  if (![response isKindOfClass:[NSHTTPURLResponse class]]) {
```
mweststrate, line 33: I'm wondering it this shouldn't return `YES` instead, given that this method returns if the response data should be stripped from the network request details, which is the case for large binary data as images, videos and zips. I can imagine we don't want to send the details fo `data:` urls to Flipper either? (see also line 50)

mark-plukkido: That makes sense. I updated this PR.

PR #1002, opened 2020-04-11
Fix pencil icon visible for non-editable fields

borisdamato:
## Summary

DataDescriptor currently shows a pencil icon next to text values starting with http/https. This is because otherwise it wouldn't be possible to edit them as a click would open the link in a web browser instead of triggering the edit mode.
Because of how this logic is implemented, it doesn't take into account whether editmode is enabled or not, thus it shows the pencil icon even when there's no need for it (custom plugins with no edit mode).

## Changelog

Fixed pencil icon visible in DataInspector for non-editable text fields

## Test Plan

I tested this locally with LayoutInspector and my own custom plugin. The pencil still appears and works in the LayoutInspector but doesn't show up anymore in my custom plugin as the data there is not editable

## Result after the fix:
**Editable (LayoutInspector)**
image

**Non-Editable (Custom Plugin)**
image


desktop/app/src/ui/components/data-inspector/DataDescription.tsx
```diff
@@ -596,17 +596,22 @@ class DataDescriptionContainer extends Component<{
...
             <>
               <Link href={val}>{val}</Link>
-              <Glyph
-                name="pencil"
-                variant="outline"
-                color={colors.light20}
-                size={16}
-                style={{cursor: 'pointer', marginLeft: 8}}
-              />
+              {editable && editIcon}
```
mweststrate, line 614: Thanks for this PR! Would you mind moving the element construction back in to the last block? With the current setup, the element will be instantiated for every string that is rendered, and not used in most cases (e.g. when it i not an url, or when the url is not editable)

borisdamato: Makes sense, good catch! Refactored in the latest commit

Issue #948, opened 2020-03-30
Question: Flipper connects to device but not application


ShawnCholeva:
Hey, I am using Flipper v0.35.0 on Mac with react-native 0.62 (just migrated from 0.61.5 following here https://react-native-community.github.io/upgrade-helper/?from=0.61.5&to=0.62.0) and am having trouble getting my application to connect to the Flipper desktop app.

In the dev console in Flipper.

```
AutoUpdateVersion.tsx:47 Error: Could not call remote method 'setFeedURL'. Check that the method signature is correct. Underlying error: Could not get code signature for running application

Underlying stack: Error: Could not get code signature for running application
    at electron/js2c/browser_init.js:6702:67
    at IpcMainImpl.<anonymous> (electron/js2c/browser_init.js:6567:27)
    at IpcMainImpl.emit (events.js:210:5)
    at WebContents.<anonymous> (electron/js2c/browser_init.js:3861:23)
    at WebContents.emit (events.js:210:5)
    at electron/js2c/browser_init.js:6705:21
    at IpcMainImpl.<anonymous> (electron/js2c/browser_init.js:6567:27)
    at IpcMainImpl.emit (events.js:210:5)
    at WebContents.<anonymous> (electron/js2c/browser_init.js:3861:23)
    at WebContents.emit (events.js:210:5)
  
user.tsx:22 Error: Feature not implemented
    at getUser (user.tsx:13)
    at _default (user.tsx:16)
    at index.tsx:51
    at Array.map (<anonymous>)
    at _default (index.tsx:51)
    at init.tsx:109
    at Object.rehydrate (persistStore.js:82)
    at _rehydrate (persistReducer.js:61)
    at persistReducer.js:90

  IOSDevice.tsx:119 Error from getpwuid_r: 102 (Operation not supported on socket)

 (123 times) server.tsx:527 Connection loop detected with iOS-iPhone Simulator-MyApp. Connected 4 times within 20s. server
```

And under the left pane, I see the following sections.

React Native
 - Logs
 - React DevTools

iPhone 11
 - Crash Reporter
 - Logs

Connecting...
  - MyApp (with a loading spinner)

It looks like the connection to the application is just not happening.  Any advice on how to fix this?

passy:
Hey @ShawnCholeva! Thanks for getting in touch. Could you click on the "Manage Plugins" button to see if there's any additional information there and check out our [Troubleshooting](https://fbflipper.com/docs/troubleshooting.html#connection-issues) page?

[...]

PR #1008, opened 2020-04-14
Update setup doc for react native swift

sunnylqm:
## Summary

If you are using swift in react native, the flipper doc has some issues:

1. Missing flags #341 which makes flipper totally ignored so that it can not connect
2. There are 2 wrong lines in `initializeFlipper` function. Need to comment them out to pass the build.

(Maybe) fixes https://github.com/facebook/flipper/issues/948

Also merge changes from https://github.com/facebook/react-native/commit/e5497ca8f6e3b240948fdbeef0ac2a710f25bb56

## Changelog

[doc] Update setup doc for react native swift

## Test Plan

Use flipper for react native without issues

### DIFF
--- a/docs/getting-started/react-native-ios.md
+++ b/docs/getting-started/react-native-ios.md
@@ -15,13 +15,34 @@ Add this code to your `ios/Podfile`:
 ```ruby
 platform :ios, '9.0'
 
-def flipper_pods()
-  flipperkit_version = '0.37.0'
-  pod 'FlipperKit', '~>' + flipperkit_version, :configuration => 'Debug'
-  pod 'FlipperKit/FlipperKitLayoutPlugin', '~>' + flipperkit_version, :configuration => 'Debug'
-  pod 'FlipperKit/SKIOSNetworkPlugin', '~>' + flipperkit_version, :configuration => 'Debug'
-  pod 'FlipperKit/FlipperKitUserDefaultsPlugin', '~>' + flipperkit_version, :configuration => 'Debug'
-  pod 'FlipperKit/FlipperKitReactPlugin', '~>' + flipperkit_version, :configuration => 'Debug'
+def add_flipper_pods!(versions = {})
+  versions['Flipper'] ||= '~> 0.37.0'
+  versions['DoubleConversion'] ||= '1.1.7'
+  versions['Flipper-Folly'] ||= '~> 2.2'
+  versions['Flipper-Glog'] ||= '0.3.6'
+  versions['Flipper-PeerTalk'] ||= '~> 0.0.4'
+  versions['Flipper-RSocket'] ||= '~> 1.1'
+  pod 'FlipperKit', versions['Flipper'], :configuration => 'Debug'
+  pod 'FlipperKit/FlipperKitLayoutPlugin', versions['Flipper'], :configuration => 'Debug'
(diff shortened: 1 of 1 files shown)


passy:
@priteshrnandgaonkar Can you take a look?

docs/getting-started/react-native-ios.md
```diff
@@ -151,12 +185,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     #if FB_SONARKIT_ENABLED
       let client = FlipperClient.shared()
       let layoutDescriptorMapper = SKDescriptorMapper(defaults: ())
-      FlipperKitLayoutComponentKitSupport.setUpWith(layoutDescriptorMapper)
```
priteshrnandgaonkar, line 154: This should work if you add the pod ` pod 'FlipperKit/FlipperKitLayoutComponentKitSupport', versions['Flipper'], :configuration => 'Debug'`

Regarding

docs/getting-started/react-native-ios.md
```diff
@@ -151,12 +185,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
       client?.add(FlipperKitLayoutPlugin(rootNode: application, with: layoutDescriptorMapper!))
       client?.add(FKUserDefaultsPlugin(suiteName: nil))
       client?.add(FlipperKitReactPlugin())
       client?.add(FlipperKitNetworkPlugin(networkAdapter: SKIOSNetworkAdapter()))
-      client?.add(FlipperReactPerformancePlugin.sharedInstance())
```
priteshrnandgaonkar, line 159: Same for here, you will have to add `pod FlipperReactPerformancePlugin` to make it work.

docs/getting-started/react-native-ios.md
```diff
@@ -43,8 +64,21 @@ def flipper_post_install(installer)
         cflags << '-DFB_SONARKIT_ENABLED=1'
       end
       config.build_settings['OTHER_CFLAGS'] = cflags
+      if (config.build_settings['OTHER_SWIFT_FLAGS'])
```
priteshrnandgaonkar, line 67: This change is just relevant for Swift projects, can we have two sections in the docs, one for Objective C and one for Swift.

priteshrnandgaonkar:
@lucasbento, adding you here, as there seems to be swift specific change which will be required in the upgrade helper.

docs/getting-started/react-native-ios.md
sunnylqm: Since there is a if condition, should do no harm to OC project right?

priteshrnandgaonkar: Even OC projects have OTHER_SWIFT_FLAGS build settings.

sunnylqm: Even so, still sounds reasonable to use one configuration for all if no actual conflicts right? Imagine one wants to turn a stock rn project into swift, how can he know he should add one more flag without any hint? (there will be no errors at all)

lucasbento:
> @priteshrnandgaonkar: adding you here, as there seems to be swift specific change which will be required in the upgrade helper.

Thank you for the ping!

I assume this will be released on react-native `0.63`, is that correct?

priteshrnandgaonkar:
Yupp, will be worth to add it in a template with a swift specific function.

PR #1018, opened 2020-04-16
Add support for deleting  a shared preference

### LINKED ISSUE #451: Discussion: Add support for deleting a shared preference
In Flipper 0.20, I can edit a shared preference item, but I can't delete one. If I delete the contents of a string preference, I end up with an empty string but I can't remove the preference item.

mzgreen:
## Summary

This change makes it possible to remove preferences. I also added a `Delete` context menu option to `DataInspector` because I needed it to implement this feature. @passy confirmed that it makes sense to add this because delete is a common action.

Fixes #451


mzgreen:
Not  sure why one of the pipelines failed - was it a timeout? Can't see anything useful in the log.

passy:
> Not sure why one of the pipelines failed - was it a timeout? Can't see anything useful in the log.

Yeah, don't worry. We've disabled these tests now as they were getting super flaky lately.

mzgreen:
Thank you  @passy ! I was worried about desktop part since I don't have much experience with React so I'm happy that ok 😃

PR #1026, opened 2020-04-17
Github action to validate Flipper and FlipperKit pod

priteshrnandgaonkar:
## Summary

This PR adds Github Action to validate the podspec file. Validation is crucial, if it podspecs don't validate then we won't be able publish pods to the trunk.
## Changelog

Add Github Action to validate FlipperKit and Flipper pod

## Test Plan

Github Action passes.

### DIFF
--- a/.github/workflows/iOS-pod-lint.yml
+++ b/.github/workflows/iOS-pod-lint.yml
@@ -1,0 +1,30 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+on: [push, pull_request]
+
+jobs:
+  lint-flipperkit_pod:
+    runs-on: macos-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/iOS-pod-lint.yml
```diff
@@ -0,0 +1,30 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
```
passy, line 3: Can you rename this to something more descriptive? "Validate Podspecs"? :)

priteshrnandgaonkar:
@passy, but validation might take a while, as it builds for all architecture, is that fine, based on the current run, it took 44 mins.

PR #1033, opened 2020-04-20
Greenkeeper/main/@testing library/react 10.0.3

mweststrate:
Created PR to investigate why the build fails on this

Solves #1032


passy:
I think that's the same error as here: https://github.com/facebook/flipper/runs/603151473

mweststrate:
Closing and recreating, as this was caused by a CI issue

PR #1083, opened 2020-04-30
[WIP]Publish pods gh

priteshrnandgaonkar:
## Summary

This PR adds a github action to publish pods on a new tag. It also tries to fix a race condition which happens when we try to publish FlipperKit pod. During the release of FlipperKit pod, we rely on the newly published Flipper Pod, and sometimes the release of Flipper Pod takes a time to get updated in the cocoapods trunk. 

This is a known issue by the Cocoapods and they will be shipping a fix to this in Cocoapods 1.10. Till then we can keep retrying for 5 times.

Ref: https://github.com/CocoaPods/CocoaPods/issues/9502#issuecomment-579486258

## Changelog

- Adds Github actions to publish cocoapods.
- Removes the Circle CI config to publish on cocoapods

## Test Plan

I tested this action by running it on this PR and it worked fine.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,25 +28,6 @@ jobs:
           command: |
             yes | sdkmanager "platforms;android-27" || true
             ./gradlew :android:assembleRelease && scripts/publish-android-release.sh
-  publish-pods:
-    macos:
-      xcode: "10.2.1"
-    steps:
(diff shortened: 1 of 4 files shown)


alloy:
@priteshrnandgaonkar Heads-up that v0.41 was not published.

PR #1086, opened 2020-05-01
Entirely control Flipper being enabled through Podfile

alloy:
## Summary

Currently user’s are being told to add a definition of the `FB_SONARKIT_ENABLED` macro and examples, including those in stock React Native templates, set this for the user by making use of a `post_install` hook in the user’s `Podfile`. This leads to confusion, fragile code [when a user’s project dir structure deviates from vanilla], and is ultimately not necessary as CocoaPods already has dedicated mechanisms to:

* specify build settings (through the `xcconfig` property);
* and selectively include certain pods only in certain build configurations (e.g. debug).

Finally, this PR also includes a commit [to fix the current builds](https://github.com/facebook/flipper/pull/1086/files#r418526812).

## Changelog

> Entirely control Flipper being enabled through inclusion in Podfile and optionally limiting to certain build configurations using the `:configuration` directive.

## Test Plan

I have built and ran the Sample application, as well as used this version of Flipper with a new RN app built from `master`.


FlipperKit.podspec
```diff
@@ -128,7 +129,8 @@ Pod::Spec.new do |spec|
   spec.subspec "FlipperKitLayoutComponentKitSupport" do |ss|
     ss.header_dir = "FlipperKitLayoutComponentKitSupport"
     ss.dependency             'FlipperKit/Core'
-    ss.dependency             'ComponentKit', '~> 0.0'
+    ss.dependency             'ComponentKit', '~> 0.30'
+    ss.dependency             'RenderCore', '~> 0.30'
```
alloy, line 133: Current versions of Flipper rely on a header that exists in a different library than `ComponentKit`, namely `RenderCore`, and _only_ exists there starting from `v0.30`.

See:
* https://github.com/facebook/flipper/pull/1086/files#diff-364ae45351222c8099cb08ea04eacdeeR13
* https://github.com/facebook/flipper/commit/d0803ecd56c7e074de7ba3eb3af35ff3641553e2#diff-364ae45351222c8099cb08ea04eacdeeR13

.gitignore
```diff
@@ -15,7 +15,6 @@ bundle.map
 **/Pods/
 **/xcuserdata/
 build/
-Podfile.lock
```
alloy, line 18: Unsure why this was being ignored? It should normally be checked in to ensure consistent builds.

iOS/Plugins/FlipperKitLayoutPlugin/FlipperKitLayoutComponentKitSupport/SKComponentLayoutWrapper.h
```diff
@@ -10,7 +10,7 @@
 #import <ComponentKit/CKComponentLayout.h>
 #import <ComponentKit/CKFlexboxComponent.h>
 #import <ComponentKit/CKOptional.h>
-#import <ComponentKit/CKVariant.h>
+#import <RenderCore/CKVariant.h>
```
alloy, line 13: See https://github.com/facebook/flipper/pull/1086/files#r418526812

iOS/Sample/Podfile
```diff
@@ -1,53 +1,25 @@
 project 'Sample.xcodeproj'
 source 'https://github.com/facebook/Sonar.git'
-source 'https://github.com/CocoaPods/Specs'
+source 'https://cdn.cocoapods.org'
```
alloy, line 3: Way faster

iOS/Sample/Podfile
```diff
@@ -1,53 +1,25 @@
...
-  #           Pod::Target::BuildType.static_library
-  #         end
-  #       end
-  #     end
-  # end
-
-  # It also adds -DFB_SONARKIT_ENABLED=1 flag to OTHER_CFLAGS, necessary to build expose Flipper classes in the header files
-  post_install do |installer|
-    file_name = Dir.glob("*.xcodeproj")[0]
-    app_project = Xcodeproj::Project.open(file_name)
```
alloy, line 39: These lines are examples of what I referred to as “fragile”.

iOS/Sample/Podfile
```diff
@@ -1,53 +1,25 @@
...
+  pod 'FlipperKit/FlipperKitReactPlugin', :path => '../../FlipperKit.podspec', :configuration => 'Debug'
+  pod 'Flipper', :path => '../../Flipper.podspec', :configuration => 'Debug'
+  pod 'Flipper-DoubleConversion', :configuration => 'Debug'
+  pod 'Flipper-Folly', :configuration => 'Debug'
+  pod 'Flipper-Glog', :configuration => 'Debug'
+  pod 'Flipper-PeerTalk', :configuration => 'Debug'
+  pod 'CocoaLibEvent', :configuration => 'Debug'
+  pod 'boost-for-react-native', :configuration => 'Debug'
+  pod 'OpenSSL-Universal', :configuration => 'Debug'
+  pod 'CocoaAsyncSocket', :configuration => 'Debug'
```
alloy, line 23: While this sample app is of course only about showcasing Flipper, I figured it might be helpful to readers to see this limiting to debug builds–and also to you when inspecting the resulting `xcconfig` files.

iOS/Sample/Sample.xcodeproj/project.pbxproj
```diff
@@ -468,41 +427,6 @@
...
-					"\"${PODS_ROOT}/Headers/Public/PeerTalk\"",
-					"-isystem",
-					"\"${PODS_ROOT}/Headers/Public/RSocket\"",
-					"-isystem",
-					"\"${PODS_ROOT}/Headers/Public/Yoga\"",
-					"-isystem",
-					"\"${PODS_ROOT}/Headers/Public/YogaKit\"",
-					"-isystem",
-					"\"${PODS_ROOT}/Headers/Public/glog\"",
-					"-DFB_SONARKIT_ENABLED=1",
```
alloy, line 504: No longer needed to specify this explicitly.

(All the other build settings never needed to be set explicitly.)

FlipperKit.podspec
```diff
@@ -85,6 +85,7 @@ Pod::Spec.new do |spec|
                                "ONLY_ACTIVE_ARCH": "YES",
                                "DEFINES_MODULE" => "YES",
                                "HEADER_SEARCH_PATHS" => header_search_paths }
+    ss.xcconfig = { "GCC_PREPROCESSOR_DEFINITIONS" => "FB_SONARKIT_ENABLED=1", "OTHER_SWIFT_FLAGS" => "-Xcc -DFB_SONARKIT_ENABLED=1" }
```
alloy, line 88: Sets the `FB_SONARKIT_ENABLED` macro for both (Obj)C and Swift targets.

alloy:
In case this is approved, any idea when a new Flipper release will be cut? It would be great to include this in an upcoming `v0.63.0` release candidate.

FlipperKit.podspec
alloy: Note that I haven’t tested this in a Swift project, as I have none; this is entirely based on the previous docs about integrating with CocoaPods into a Swift project.

FlipperKit.podspec
alloy: Note that I’m unsure how this has been building previously, as it seems that that change already exists since Flipper v0.38?

alloy:
I don’t think any of the CI failures are related to my changes, but let me know if otherwise!

iOS/Sample/Podfile
```diff
@@ -1,53 +1,25 @@
...
+  pod 'ComponentKit', '~> 0.30'
```
priteshrnandgaonkar, line 24: Why did you add this ?

iOS/Sample/Podfile.lock
```diff
@@ -0,0 +1,130 @@
+PODS:
```
priteshrnandgaonkar, line 1: I agree Podfile.lock should be committed. I am working on it and trying to automatically update Podfile.lock when a release is made. Till then can you remove the podfile.lock and also add it in gitignore.

priteshrnandgaonkar:
New flipper release gets cut every Tuesday. So you will have to wait for next tuesday.

FlipperKit.podspec
priteshrnandgaonkar: We have a [SampleSwift](https://github.com/facebook/flipper/tree/master/iOS/SampleSwift) sample project too. Try to run it and make this changes over there too, if possible.

.gitignore
priteshrnandgaonkar: Can you remove this. I have mentioned the reason in the comment on the Podfile.lock

iOS/Sample/Podfile
alloy: To illustrate to readers that while ComponentKit is a transitive dep of the ComponentKit plugin, the app itself directly relies on it too and thus it should be listed explicitly.

FlipperKit.podspec
alloy: Doh, I entirely missed that 🤦 Will do!

.gitignore
alloy: Done 👍

FlipperKit.podspec
alloy: Done 👍

alloy:
Addressed your feedback 👍

priteshrnandgaonkar:
@alloy Thanks for addressing it, bytw there is a tutorial app as well which uses the references the released Flipper version instead of referencing the local one. Would you like to change that as well, otherwise I can change it. 

Apologies for not mentioning abt that before.

alloy:
> @alloy Thanks for addressing it, bytw there is a tutorial app as well which uses the released Flipper version instead of referencing the local one.

We’d need to wait till a new version is released to ensure it actually works, so perhaps you could follow-up with that once the release is out? I can do so too, but it might not be until the end of the week at the earliest.

iOS/Sample/Podfile
priteshrnandgaonkar: Unfortunately, I will have to remove this, as our internal test is facing an issue which is mentioned [here](https://github.com/CocoaPods/CocoaPods/issues/9270) and [here](https://github.com/CocoaPods/CocoaPods/issues/9731)

PR #1092, opened 2020-05-01
Fix tests failing on github

nikoant:
Summary: Fixed tests failing on github. Our custom transformation for 'fb-stubs' is not applied for tests and jest.mocks in particular, and therefore we need to try mock both 'fb' and 'fb-stubs' object depending on where the test is running.

Differential Revision: D21352524


nikoant:
This was fixed by another PR

PR #1118, opened 2020-05-04
Fix link to docs

tchayen:
## Summary

Link to documentation in the README is broken. I have replaced it with a working one.

## Changelog

Fixed link to the docs in README

## Test Plan

Check out both links in the browser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ Older yarn versions might show an error / hang with the message 'Waiting for the
 
 # Documentation
 
-Find the full documentation for this project at [fbflipper.com](https://fbflipper.com/docs).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -121,7 +121,7 @@ Older yarn versions might show an error / hang with the message 'Waiting for the
 
 # Documentation
 
-Find the full documentation for this project at [fbflipper.com](https://fbflipper.com/docs).
+Find the full documentation for this project at [fbflipper.com](https://fbflipper.com/docs/features/index/).
```
passy, line 124: Hm, I think we would want this to point just to `https://fbflipper.com/` in this case.

tchayen: I have updated the link

PR #1182, opened 2020-05-23
RegExp doesn't need underscore with \w class

zusinShinpei:
## Summary
Character class '\w' has alphabets, digits, underscore. so we don't need to write '_' character with '\w' class.

### DIFF
--- a/desktop/doctor/src/index.ts
+++ b/desktop/doctor/src/index.ts
@@ -38,24 +38,28 @@ export type Healthcheck = {
   key: string;
   label: string;
   isRequired?: boolean;
-  run: (env: EnvironmentInfo) => Promise<HealthchecRunResult>;
+  run: (env: EnvironmentInfo) => Promise<HealthCheckRunResult>;
 };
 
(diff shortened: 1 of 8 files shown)


passy:
Great find! Thank you!

desktop/doctor/src/index.ts
```diff
@@ -38,24 +38,28 @@ export type Healthcheck = {
   key: string;
```
passy, line 38: The changes here seem unrelated. Could you split this off into a separate PR if you want to submit them? :)

desktop/doctor/src/index.ts
```diff
@@ -256,7 +259,7 @@ export async function runHealthchecks(): Promise<
 
 async function tryExecuteCommand(
   command: string,
-): Promise<HealthchecRunResult> {
+): Promise<HealthCheckRunResult> {
```
passy, line 262: This is a great catch, too, but it would be ideal if those were individual submissions.

Issue #1183, opened 2020-05-23
Database plugin doesn't work with Room + requery/sqlite-android


aartikov:
## 🐛 Bug Report
Database plugin shows outdated data when it is connected to Android application that uses [Room](https://developer.android.com/training/data-storage/room) and [requery/sqlite-android](https://github.com/requery/sqlite-android). Moreover, the plugin prevents Android app to save data to a file.

## To Reproduce
1. Build and run the application [https://github.com/aartikov/flipper-database-bug](https://github.com/aartikov/flipper-database-bug)
2. Connect to it with Flipper Desktop and open word_table.
3. Add some data to the table by pressing (+) button  in the Android app.
4. Press Refresh button in Flipper Desktop. It will display outdated data. 
5. Press Back in the Android app and swipe it away from recent task list to kill it.
6. Start the Android app again. It will not show the data that was added on the step 3.

You can try to remove line `.openHelperFactory(RequerySQLiteOpenHelperFactory())` in my sample. In that case everything will work fine.
Also, if you will repeat steps 3, 5, 6 without Flipper Desktop running, the data will be saved correctly.

## Environment
- Flipper 0.43.0
- Android 9
- Windows 10 or Ubuntu 20.04 LTS

aartikov:
I figured out that I have to implement custom `SqliteDatabaseConnectionProvider` that returns requery implementation `io.requery.android.database.sqlite.SQLiteDatabase`. To make it possible this change is required #1196

PR #1196, opened 2020-05-26
Use interface SupportSQLiteDatabase in Android Databases Plugin

aartikov:
## Summary
This change will allow to use various SQLiteDatabase implementations: standard Android implementation, [requery/sqlite-android](https://github.com/requery/sqlite-android) and so on. See issue #1183 

## Changelog
Android Databases Plugin: `SqliteDatabaseConnectionProvider` returns `SupportSQLiteDatabase` instead of `SQLiteDatabase`.

## Test Plan
Check that Databases Plugin shows correct data on the sample Android application.


passy:
This looks great! Thanks for doing this. I'll import this for internal review.

aartikov:
@passy I have added some fixes. Please, import it again.

passy:
@aartikov Really sorry for the wait. Do you know where `androidx.sqlite.db.framework.FrameworkSQLiteDatabaseWrapping` is coming from or when it was added? This is failing to resolve internally.

aartikov:
@passy It is a part of the pull request https://github.com/facebook/flipper/pull/1196/files#diff-6deeccf2669d9a45af844801f3979441
I placed it to `androidx.sqlite.db.framework` to get access to package-private class `FrameworkSQLiteDatabase`.

passy:
@aartikov Oh, I didn't see that at first. Sorry, I'm afraid we can't ship with that. This doesn't work well with our build system and it's generally not a good form to publish artifacts into someone else's namespace.

passy:
@aartikov Do you think reflection would be an option here at all? It may seem a little ugly, but it would help with compatibility here.

aartikov:
@passy Yes, we can use reflection here. I have pushed this change.

aartikov:
@passy could you please review it?

passy:
@aartikov Apologies for the delay! I'm a little swamped right now, but I'll take a look this week.

likern:
@aartikov Hello, not sure my problem is the same.

But when I open flipper and try to refresh database - it looks like database **is completely deleted** (I suggest by flipper itself).
Also I can sometimes see database itself, but never tables (emtpy popup list).

I use React Native with https://github.com/breeffy/mobbly/commit/86813c1f4bab7e1b1719bea0aa86b416c5937391 which itself uses https://github.com/requery/sqlite-android.

I tried your patch (I hoped it's related, but not sure). Didn't help to fix my problem.

PR #1200, opened 2020-05-27
Updated links in documentation

berdimyradov:
- Links are corrupted
- Updated them


mweststrate:
I just tried these links on https://fbflipper.com/docs/getting-started/react-native and they seem to be working fine. Also note that links should be relative not absolute, as otherwise it will be impossible to verify changes before publishing them to production.

Issue #262, opened 2018-09-06
[iOS] Support Physical iOS devices


priteshrnandgaonkar:
Currently for iOS flipper only supports debugging through simulators but not on physical devices. The issue is that we were not able to copy generated certificates in the mobile data partition where app can access it. More information about certificate exchange in Flipper is [here](https://fbflipper.com/docs/establishing-a-connection.html).

OleksiyA:
One idea how we can do this: in iOS SDK code provide hook point for developers to implement logic "should connection from host XYZ be allowed?". Some developers would allow all connections and rely on WiFi router configuration to limit outside connections.

One more way: Use libmobiledevice to write files to mobile device directly (https://github.com/libimobiledevice/ifuse). This would require App to enable iTunes file sharing. Also iPhone have to be attached to computer by USB cable.

alecgorge:
Could copying the generated certificate into the app bundle as a build step be another viable avenue?

digoreis:
Guys, we have a news about this ? 
My suggestion is in iOS case use [multipeerconnectivity.](https://developer.apple.com/documentation/multipeerconnectivity).

jknoxville:
The above suggestions could work. As Pritesh mentioned originally the only missing piece is deploying the files into the app container. That could indeed be done either as a build step, or by implementing the push and pull methods of this class: https://github.com/facebook/flipper/blob/master/src/fb-stubs/iOSContainerUtility.js by any means necessary.

There's work ongoing to open source the solution we're using for this internally, but I don't expect that to be ready until at least May of this year.

digoreis:
@jknoxville I did a small study here.  Thanks for the indications 😄 

For to use MultipeerConnectivity, we  have a small problem with the init of `FlipperClient::_cppClient`.  **Problem** is that a discovery DNS protocol like *Bonjour*, **always** an async operation. 

For the solution this, I research to create a FlipperDiscovery. This entity will create a connect, get all information and copy the certificate files. Then, to start `FlipperClient` with IP and files correct. 

I'm working in a test solution but very slow 🐌 , I need to remember Obj-C and C++. 😅

bizzguy:
I'm wondering if there is any updated on when iOS physical devices will be supported.  Thanks.

GioLogist:
I'm having a difficult time getting iOS Simulator to work as well. 

Initialized via 

```bash
react-native init reactNative610Rc --version react-native@0.61.0-rc.0
```

[...]

PR #1217, opened 2020-06-02
Master PR: Enable physical iOS devices

jknoxville:
## Summary

This is to fix https://github.com/facebook/flipper/issues/262

Currently a work in progress, there are various fixes to make before the feature can be enabled by default.

## Changelog

Now physical iOS devices are supported in open-source Flipper.

## Test Plan

* Verified connection is made and network inspector works.


jknoxville:
Closing this PR, because I'm just adding commits to master as I go instead, and we're nearly there.
This commit shows how to enable the feature yourself if you need to before the feature is complete, but won't be merged into production.

Issue #1206, opened 2020-05-30
Network plugin mock data not maintained after plugin switch.


mgoyanes:
Hello.
Let me start by thanking @qichuan for this amazing option added to the network plugin. I use and abuse of the mock option.

Don't know if this is a bug or not but after adding mock data and then switching to another plugin or restarting the Android app, the mock data is lost thus I have to insert it again.

As a user, I was expecting the mock data being persisted if I switch plugin or restart my app thus not needing to insert the mock data every time one of the described situations occurs.

## To Reproduce

1. In the network, plugin click the mock button and add mock data.
2. Switch to another plugin (databases for example)
3. Go back to the network plugin and press the mock button.
4. No mock data is available

## Suggestion
Another thing that would be great was the place where we insert the mock data had json code highlighting.
Also, the pre-filled `/` on the URL for me is unnecessary. I always have to delete it or double click it to add my URL.

## Environment
Using Flipper on Android. Flipper client version 0.44 on Mac.

qichuan:
Thanks for the feedback, let me see how I can make this plugin work better

mgoyanes:
Hello.

Another thing. The close button is not fully visible.

![image](https://user-images.githubusercontent.com/5049948/83518078-e6788680-a4d1-11ea-97e9-2f01301421db.png)

mgoyanes:
I know that this comes late, but it would be great if we could toggle on or off the mock for a specific URL.

Currently, if in a session we don't want to mock a response, we have to delete the mock.

PR #1218, opened 2020-06-02
Persist the Network mock data

qichuan:
## Summary

Persist the Network mock data, so it can survive after switching the plugins

Closes https://github.com/facebook/flipper/issues/1206
## Changelog

- Persist the Network mock data
- Remove the pref-filled '/' when creating a mock route

## Test Plan

Screenshot 2020-06-02 at 11 45 15 PM

### DIFF
--- a/desktop/plugins/network/index.tsx
+++ b/desktop/plugins/network/index.tsx
@@ -209,15 +209,16 @@ export default class extends FlipperPlugin<State, any, PersistedState> {
 
   init() {
     this.client.supportsMethod('mockResponses').then((result) => {
-      const routes = JSON.parse(localStorage.getItem(LOCALSTORAGE_MOCK_ROUTE_LIST_KEY) || '{}')
+      const routes = JSON.parse(
+        localStorage.getItem(LOCALSTORAGE_MOCK_ROUTE_LIST_KEY) || '{}',
+      );
       this.setState({
         routes: routes,
         isMockResponseSupported: result,
         showMockResponseDialog: false,
         nextRouteId: routes.length,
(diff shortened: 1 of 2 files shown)


passy:
Brilliant. Thanks for working on that and adding this so quickly. :)

mgoyanes:
Thanks @qichuan

desktop/plugins/network/index.tsx
```diff
@@ -206,14 +208,18 @@ export default class extends FlipperPlugin<State, any, PersistedState> {
   }
 
   init() {
-    this.client.supportsMethod('mockResponses').then((result) =>
+    this.client.supportsMethod('mockResponses').then((result) => {
+      const routes = JSON.parse(
+        localStorage.getItem(LOCALSTORAGE_MOCK_ROUTE_LIST_KEY) || '{}',
```
cekkaewnumchai, line 213: You should use `persistentState` and `setPersistentState` from props. This can preserve the change for each session.

qichuan: I use `localStorage` to persis the mocks, so they can survive even if the Flipper app is restarted

cekkaewnumchai: Then, I'll let @mweststrate stamp.

mweststrate: Note that using `localStorage` means that all instances across different devices / applications will have the same mocks available (which can be seen both as a feature and a bug I guess :-P)

mweststrate:
(sorry for the late follow up!)

PR #1227, opened 2020-06-05
Updated dependencies to latest versions and updated filenames in FlipperKit

### LINKED ISSUE #1226: Question: Pods won't install on FlipperKit, iOS
When running `pod install` I get the following:
```
pod install
Analyzing dependencies
[!] CocoaPods could not find compatible versions for pod "Flipper-RSocket":
  In Podfile:
    Flipper (from `../Flipper.podspec`) was resolved to 0.45.0, which depends on
      Flipper-RSocket (~> 1.1)

    Flipper-RSocket (~> 0.10)

Specs satisfying the `Flipper-RSocket (~> 0.10), Flipper-RSocket (~> 1.1)` dependency were found, but they required a higher minimum deployment target.
```

jbfitb:
## Summary

I was unable to build `FlipperKit` for iOS. It looks like there's a dependency mismatch. See https://github.com/facebook/flipper/issues/1226

## Changelog

 - Updated Podfile dependencies to latest versions
 - Added a default scheme for building
 - Updated the iOS platform version to iOS 10.0 from 9
 - Updated filenames to match the new filenames

## Test Plan

Was able to successfully build:

Fixes #1226 
Screen Shot 2020-06-05 at 10 49 54 AM

Screen Shot 2020-06-05 at 10 50 25 AM

### DIFF
--- a/iOS/FlipperKit.xcodeproj/project.pbxproj
+++ b/iOS/FlipperKit.xcodeproj/project.pbxproj
@@ -8,7 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		027369E09714D434B8C421F6 /* libPods-FlipperKit.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 0D8E34F2706E57CC3C186A67 /* libPods-FlipperKit.a */; };
-		53915CB22152E3040090EEA6 /* SonarResponder.h in Headers */ = {isa = PBXBuildFile; fileRef = 53915C972152E3030090EEA6 /* SonarResponder.h */; };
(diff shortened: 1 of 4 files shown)


priteshrnandgaonkar:
FlipperKit.xcodeproj has been neglected since a long time, maybe because, most of the times we end up working on Sample app xcodeproject which locally links to the FlipperKit project files and thus allowing us to quickly iterate and test those changes in the sample app as well.

I would like to know what workflow you were following to work on FlipperKit.xcodeproj. Otherwise we can remove it altogether.

jbfitb:
@priteshrnandgaonkar See #1226 for discussion. Goal is to build and archive so that I can use with carthage.

priteshrnandgaonkar:
Cool, sounds good to me then.

PR #1256, opened 2020-06-15
RegExp doesn't need underscore with \w class

zusinShinpei:


passy:
Hey @zusinShinpei! Sorry for the delay but I've noticed that this breaks a few tests and that makes me a little uneasy. There seem to be some replacements which aren't quite equivalent. 

In the crash reporter, the paths for instance don't get parsed correctly anymore:

Screenshot 2020-06-23 17 04 02

passy:
Closing this as @mweststrate's commit contained most of the changes.

Thanks for bringing this up, @zusinShinpei!

PR #1268, opened 2020-06-15
escape app name to account for special characters in name (closes #1252)

### LINKED ISSUE #1252: Plugins not enabled in iOS simulator
## 🐛 Bug Report
After upgrading to 0.46.0, the Flipper plugins no longer are able to be enabled.
Flipper_and_New_Issue_·_facebook_flipper

## To Reproduce
Launch Flipper, Launch iOS app in simulator, try to enable plugins.

## Environment
iOS 13.5
Simulator 11.5
React Native 0.62.1
Flipper 0.46.0

Plugins installed and supported:
Flipper

casconed:
## Summary

App Names can contain special characters. Escape them (and unescape later) so that device ID strings don't break. 

Simpler is better - thanks to @mweststrate for keeping me out of the rabbit hole
## Changelog

Escape app name to account for special characters in name
## Test Plan
Added test to test character escaping: `cd desktop && yarn test -- -i app/src/utils/__tests__/clientUtils.node.tsx`

closes #1252

### DIFF
--- a/desktop/app/src/utils/__tests__/clientUtils.node.tsx
+++ b/desktop/app/src/utils/__tests__/clientUtils.node.tsx
@@ -48,3 +48,21 @@ test('client id deconstruction error logged', () => {
   });
   expect(consoleErrorSpy).toHaveBeenCalledTimes(1);
 });
+
+test('special characters in app name handled correctly', () => {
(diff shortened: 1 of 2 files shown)


jknoxville:
Thanks for doing this!

PR #1269, opened 2020-06-16
[flipper-pkg] When init, search for first level subdirectory

usrbowe:
## Summary

When running `flipper-pkg init path/plugin`, this allows to skip the `Do you wanna add this path to watching in Flipper config?`, if the path is direct first level subdirectory of any added path from Flipper `pluginPaths: []`

## Changelog

- Improve matching for watched pluginPaths when initialising new plugin

## Test Plan

1. Create folder: `~/flipper-plugins`
2. Add this folder to Flipper config `pluginPaths: ['~/flipper-plugins']`
3. Create new subfolder `test-plugin`
4. Run `flipper-pkg init test-plugin`
5. It should not ask to add this folder to pluginPaths

To verify it only works for first level subdirectory, create another folder: `/deep/test-plugin2` and re-run steps with it.

This should also keep working if we specify exactly the path for the plugin itself.


jknoxville:
Good find, this makes sense to me.

Issue #1280, opened 2020-06-20
Network Mock Dialog crashes when 'Add Route' button is pressed


qichuan:
## 🐛 Bug Report

Network Mock Dialog crashes when 'Add Route' button is pressed

## To Reproduce

1. Open Network Plugin
2. Press on the Mock button
3. Press Add Route button
4. The Network Mock Dialog crashes
<img width="1026" alt="Screen Shot 2020-06-20 at 4 05 00 PM" src="https://user-images.githubusercontent.com/410850/85196870-cccba180-b30f-11ea-9dc6-0f1139ac5d9b.png">

## Environment

Latest master branch, git hash: 4ac9dc9edadf4fb0be2d6b45345a1af4552e53e4

PR #1281, opened 2020-06-20
Fix the network plugin crashes when Add Route button is pressed 

qichuan:
## Summary

Fix Network Mock Dialog crashes when 'Add Route' button is pressed  https://github.com/facebook/flipper/issues/1280


passy:
Cheers!

PR #1297, opened 2020-06-26
Fix link to React Native setup instructions

sonicdoe:


passy:
I've found a bunch more issues like the one you uncovered and fixed them all together. Thanks again for flagging!

sonicdoe:
This pull request isn’t about the relative link but about changing the link from [getting-started/index/#setup-your-react-native-app](https://fbflipper.com/docs/getting-started/index/#setup-your-react-native-app) to [getting-started/react-native](https://fbflipper.com/docs/getting-started/react-native). The former link’s anchor doesn’t exist and the latter is arguably a better destination for React Native setup instructions.

If you’d like, you can reopen this pull request and I’ll update it to the new `<Link>` format.

Issue #1300, opened 2020-06-26
Flipper leaks activities that were created but not paused.


pyricau:
## 🐛 Bug Report

ApplicationWrapper.mActivities ([ApplicationWrapper source](https://github.com/facebook/flipper/blob/master/android/src/main/java/com/facebook/flipper/plugins/inspector/ApplicationWrapper.java
)) is a list of weak refs to activities. They get added in onCreate():

```
  @Override
  public void onActivityCreated(Activity activity, Bundle savedInstanceState) {
    mActivities.add(new WeakReference<>(activity));
    notifyListener();
  }
```

and removed in onPause:

```
  @Override
  public void onActivityPaused(Activity activity) {
    if (activity.isFinishing()) {
      final Iterator<WeakReference<Activity>> activityIterator = mActivities.iterator();

      while (activityIterator.hasNext()) {
        if (activityIterator.next().get() == activity) {
          activityIterator.remove();
        }
      }

      notifyListener();
    }
  }
```

That behavior is incorrect: if an activity is created and destroyed without being paused, the weak ref will never be removed from the list. This behavior happens when you call finish() from Activity.onCreate(), in which case there's no pause.

I've been investigating some activity memory leaks that seem to be only held by weak refs.

A key issue here is that calling `ApplicationWrapper.getActivityStack()` (ie to refresh the view hierarchy for flipper) repeatedly will reactivate those weak refs and prevent garbage collection.

pyricau:
Depending on what getActivityStack() is used for, there can be different fixes:

1. You could move the mActivities cleanup to onActivityDestroyed(). That's probably you best bet, though it does mean activities will now still show up after being paused. That seems reasonable though, as a paused activity still has a window, a view hierarchy, and can still be drawing stuff on the screen: when the top of the activity stack has transparent pixels, the activity below it is paused (no touch events) but not stopped (no drawing)
2. You could add the activities back in onResume, and remove them again onDestroy(). That way you'll get symmetry of calls and you the activities won't be in the list while paused (not sure why you'd want that but maybe you do)

pyricau:
Activity I realized another thing: the weak refs are removed only if `activity.isFinishing()` is true. So on every config change (e.g. rotation) isFinishing() will be false and the weak refs are never cleared.

PR #1301, opened 2020-06-26
Fix ApplicationWrapper activity weakRef leak

pyricau:
Fix for #1300

The weak refs were not being cleared in two cases:

- On config changes, isFinishing() would be false in onPause()
- When calling finish() from Activity.onCreate(), onPause() isn't guaranteed to be called.


pyricau:
Thanks for adding tests @facebook-github-bot ;)

passy:
@pyricau Haha. Yes, it's a good bot. :P

Issue #1303, opened 2020-06-27
Cannot build Android Release with react-native-flipper 0.48.0+


TheAlmightyBob:
## 🐛 Bug Report
react-native-flipper 0.48.0 breaks building Android in Release, presumably due to https://github.com/facebook/flipper/commit/03b231703ba1e609d8b5d31d71d07c57be1f0b29

Version 0.47.0 did not have this problem, although it had the problem in #1274, which the above change was trying to address.

![image](https://user-images.githubusercontent.com/7707864/85933052-5d156380-b887-11ea-9f64-dc8fc55de028.png)

## To Reproduce
Very simple.
1. `npx react-native init Test`
2. `cd Test`
3. `yarn add react-native-flipper`
4. Open project in Android Studio
5. Set Build Variant to "release"
6. Try to build project

The build will fail, but if you change the version of `react-native-flipper` from 0.48.0 to 0.47.0, then it will succeed.

## Environment
Since this is a build failure with a default-initialized React Native project, the only variable being react-native-flipper itself, I can't think of any relevant environment info to add, but feel free to ask if you think there's something I've missed.

ObidosDev:
Same stuff

ObidosDev:
May be need to create debug folder and return module for auto linking by `BuildConfig.DEBUG` value and with reflection return FlipperModule instance if debug is true. (Very the same as for `initializeFlipper` from installation part)

Because root cause is auto generated `PackageList.java`
```
public ArrayList<ReactPackage> getPackages() {
    return new ArrayList<>(Arrays.<ReactPackage>asList(
      new MainReactPackage(mConfig),
      new FlipperPackage(),  // <- it is here in release 
...
```


```
if (BuildConfig.DEBUG) {
            try {
                Class<?> flipperModuleClass = Class.forName("com.facebook.flipper.reactnative.FlipperModule");

                return Collections.<NativeModule>singletonList((NativeModule)
                        flipperModuleClass
                                .getMethod("init", ReactApplicationContext.class)
                                .invoke(null, reactContext));

            } catch ...
```

```
public class FlipperModule extends ReactContextBaseJavaModule {

    public static final String NAME = "Flipper";

    public static FlipperModule init(ReactApplicationContext reactContext) {
        return new FlipperModule(FlipperReactNativeJavaScriptPluginManager.getInstance(), reactContext);
    } ...
```

Hope it make sense 🧑‍💻

ObidosDev:
[...]

Issue #1274, opened 2020-06-18
Flipper libraries get included in release builds on Android projects using react-native-flipper


TPXP:
## 🐛 Bug Report
When adding `react-native-flipper` (required by the [redux plugin](https://www.npmjs.com/package/redux-flipper) for example) to a project, Android release builds get bloated with Flipper binaries.

Impact: bundle size significantly increased (+3MB download, +6MB install size)

## To Reproduce
```sh
npx react-native init MyApp
cd MyApp
yarn add react-native-flipper
cd android
./gradlew bundleRelease # assembleRelease will also work (but you get an APK instead)
```

For example, `libflipper.so` is included (it also comes with unneeded `libevent`):
```sh
$ cd app/build/outputs/bundle/release
$ unzip app-release.aab
$ find . | grep flipper
./base/lib/armeabi-v7a/libflipper.so
./base/lib/x86/libflipper.so
./base/lib/arm64-v8a/libflipper.so
./base/lib/x86_64/libflipper.so
```

## Environment
React-Native 0.62.2
Android app
Looks ok on iOS (same bundle sizes with or without `react-native-flipper` - at least when using 0.47.0)

passy:
Oh, that's interesting. Agreed that it definitely shouldn't be included in release builds. Do you know how this is handled for other debug-only dependencies with native components? I'm not sure how we would declare this as part of the `react-native-flipper` package.

TPXP:
Well, to be honest that's the only debug dependency I have in my project, so I can't really help on this point. Do you know of any other debugging tools for react native?

FDiskas:
Should be reopened

mweststrate:
@FDiskas given #1325, preferably give some details / reproduction on why it should be reopened. Your comment as it stands is inactionable.

PR #1325, opened 2020-07-01
Fix: Release builds on android

b3coded:
## Summary

After `react-native-flipper` 0.48.0 I cannot build releases on android (I think that passed on CI tests because the example was running an older version of this package, 0.47.0).
So I moved all the `android/src/main` content to `android/src/debug` because we will not use Flipper in another Build Variant

## Changelog

I moved all content from `react-native-flipper/android/src/main` to `react-native-flipper/android/src/debug`.

Probably solves #1303 

## Test Plan

Maybe create a custom CI script to verify if Flipper deps are present on Release Builds, [something like this](https://github.com/facebook/flipper/issues/1274#issue-641197153)


passy:
You're a star! Thank you.

TheAlmightyBob:
I applied this change locally and the build still failed due to `PackageList` being unable to find `FlipperPackage`

b3coded:
> I applied this change locally and the build still failed due to `PackageList` being unable to find `FlipperPackage`

@passy can I add one more commit? I forgot to add a `FlipperPackage.java` inside `main` folder.
Will solve this problem

b3coded:
@passy The error still persists, I asked on ReactNative CLI repo [here](https://github.com/react-native-community/cli/issues/1211) but the answer was not positive :disappointed:

PR #1329, opened 2020-07-03
Fix release builds on Android, take 2

### LINKED ISSUE #1303: Cannot build Android Release with react-native-flipper 0.48.0+

TheAlmightyBob:
## Summary

#1325 made progress but did not quite fix #1303. Android Release builds would still fail because `PackageList` depended on `FlipperPackage` and couldn't find it in release builds. There doesn't appear to be a clear way to set up a debug-only package with React Native's Android autolinking (see https://github.com/react-native-community/cli/issues/1211), so this change creates an empty `FlipperPackage` to satisfy that dependency, while still avoiding bundling the Flipper library into the release build.

(manual linking also works, but seems less-desirable for consumers of the library)

## Changelog

Add an empty `FlipperPackage` to the `release` folder to satisfy the reference put in `PackageList` by the autolinker for release builds.

## Test Plan

Successfully built a React Native Android Release app from Android Studio using this library and verified that the APK still did not contain Flipper. Built debug and verified that Flipper still worked as before.


passy:
Thanks a lot for looking into this further, @TheAlmightyBob! Would you mind taking care of the CLA so I can import this?

b3coded:
@passy can you only import this PR if he signs the CLA? Because the `Flipper 0.50.0` was released with this bug

TheAlmightyBob:
Apologies for the delay. I have not forgotten about this, I just hadn't anticipated the CLA complication. Still working on that..

TheAlmightyBob:
@TPXP Thanks for reporting! I thought I tested that scenario, but obviously missed something since I just now repro'd the duplicate class error. Updated.

alecpo:
If we install react-native-flippe as a dev dependency, would it work?

b3coded:
@alecpo I had this idea but it's related to React Native CLI repo, my idea:

DevDependencies that use native code should only be implemented on Debug Builds, it would be awesome

TheAlmightyBob:
Thanks for everybody's patience

imranMnts:
Can we have a status please?

TheAlmightyBob:
@passy The CLA has been signed and this should be ready to go.

ryanbabbly:
Is there a plan to merge this soon? Because I have this problem all the time when building locally

rikur:
This would be very helpful to have.

ryanbabbly:
For everyone else tired of editing their `react-native-flipper` directly then losing it next time you run yarn you can copy the contents of the new file in this PR and put it in your own code under this path:

`android/app/src/release/java/com/facebook/flipper/reactnative/FlipperPackage.java`

The app will build and whenever this is finally merged you can just delete that file/path.

thallada:
Or, you could copy the file in this PR into your local node_modules and create a custom patch with [`patch-package`](https://github.com/ds300/patch-package) (patch-package automatically applies it on npm post-install)

mweststrate:
Fix looks correct, should be part of next release!

FRizzonelli:
I need to build an APK in release, but I don't understand where to put the file. Should it be inside node_modules/react-native/react-native-flipper or directly into react-native-flipper/ or in my android project? Tried all combination, downgraded to 0.47, upgrade to 0.53, copy pasted the file in every location that make sense to me but nothing. How are we supposed to use this PR? Thanks in advance.

mweststrate:
We expect a release of react-native-flipper tomorrow, but otherwise in your
node_modules/react-native-flipper indeed.

FRizzonelli:
@mweststrate ok thank you :) One last question, if you fix the error tomorrow on flipper, we also need to wait for the next release of react-native, am I correct? (from 0.62 flipper should be shipped with RN without manually adding it in package.json)

TheAlmightyBob:
@FRizzonelli This library is separate from what's included with React Native. You can use Flipper with React Native 0.62 without react-native-flipper. You need this library only (AFAIK) for using custom plugins (e.g. Reactotron, redux inspectors, async storage inspectors, etc..)

Unless there's some breaking change that depends on a new version of React Native, you shouldn't need to wait for that.

shubhamdeol:
@mweststrate waiting for new version release. 👍

mweststrate:
The change has been released as part of react-native-flipper@0.54.0. Thanks for the PR @TheAlmightyBob !

PR #1407, opened 2020-07-27
Updated CODE_OF_CONDUCT.md

chaytanyasinha:
Added full stops in the end of the sentences.

## Summary

Added full stops in the end of the sentences.


passy:
Thanks, but we're using a template for this. I believe the lack of periods there is to ensure that link detection works correctly in all setups.

chaytanyasinha:
> Thanks, but we're using a template for this. I believe the lack of periods there is to ensure that link detection works correctly in all setups.

Thanks for the knowledge

PR #1429, opened 2020-08-01
- make use of grave accents in sql queries 

### LINKED ISSUE #1426: SQL Table and Column names need to be string literals to prevent failed UPDATE statements
## 🐛 Bug Report

Having a column name *order* will cause errors when editing row data in the Flipper desktop app

## To Reproduce

Create a table that has a column named *order* and then edit that field using the desktop app

Screenshot from 2020-07-31 01-39-53

it will produce a query similar to this `UPDATE labels_table SET order=0 WHERE id=4`, which will fail to execute

Screenshot from 2020-07-31 01-41-50

No sure why this is happening or if it's a Flipper or sqlite issue, seems like the program is expecting the ORDER BY clause if *order* is not in quotes.

wrapping *order* in single quotes or grave accents and manually running `UPDATE labels_table SET 'order'=0 WHERE id=4` in the SQL query executor is currently the workaround.

## Environment

using:
Flipper Desktop 0.51.0
Android 0.51.0

lmj0011:
## Summary

closes #1426

### DIFF
--- a/desktop/plugins/databases/UpdateQueryUtil.tsx
+++ b/desktop/plugins/databases/UpdateQueryUtil.tsx
@@ -58,9 +58,9 @@ export function constructQueryClause(
           ? `'${val.value.replace(/'/g, "''")}'`
           : `${val.value}`;
       if (idx <= 0) {
-        return `${key}=${valueString}`;
+        return `\`${key}\`=${valueString}`;
       } else {
-        return `${clauses} ${connector} ${key}=${valueString}`;
+        return `${clauses} ${connector} \`${key}\`=${valueString}`;
       }
     },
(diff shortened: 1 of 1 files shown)


lmj0011:
This is going to need to be applied to all constructed SQL statements. Came across this bug again when clicking the column tab to reorder rows

Peek 2020-08-01 23-54

lmj0011:
@passy 

Another maintainer/contributor can take it from here. I'm unsure when I'll be back to using Flipper again.

passy:
@lmj0011 Thanks for letting us know!

mweststrate:
I'll pick this one up

Issue #1423, opened 2020-07-30
Discussion: Allow to mock response status code in request mocking feature


gildor:
Existing request mocking feature doesn't support providing custom HTTP status code, it always returns 200, which is quite a big limitation, the testing HTTP statuses is one of the most important features of request mocking

bizzguy:
I agree with the request to add return code.  Also, I would also like to see DELETE added as a type.  Mocking is really an AWESOME addition to the plugin!

bizzguy:
One more thing.  If the user has selected (highlighted) a network request when the Mock button is clicked, it would be useful if the current highlighted request was populated in the data section.  Some of our requests are really long and this would save us time rather than having to copy and paste from the original request.

bizzguy:
A few more things.  It appears that "export" does not export the mock definitions to the "FlipperExport.flipper" file.

Also, having a way to temporarily disable the mock without deleting the definition would be useful.

bizzguy:
Just submitted a PR to allow for specifying status code on response.  Would be interested in feedback from anyone on this thread.  Thanks. 

https://github.com/facebook/flipper/pull/1441

bizzguy:
The PR to fix this issue has been merged so this issue can now be closed.

bizzguy:
This issue can be closed.  The functionality to add response status code to mocking was added in PR https://github.com/facebook/flipper/pull/1447

PR #1441, opened 2020-08-05
Add HTTP Status Code to mock response for Route

bizzguy:
## Summary

Mock requests can currently only return an HTTP Status Code of 200.  This is not sufficient to take full advantage of the mocking feature.  It would be more useful to specify a status code for the response.  Then not only can the original request be tested, but failure of the call could be tested as well.

This is described in Issue #1423 [https://github.com/facebook/flipper/issues/1423]

## Changelog

Allow user to change response code for a mock request

## Test Plan

Here is a video demonstrating the change
[Uploading response-status-code.mp4.zip…]

Here is a screenshot showing where the new HTTP status code is entered for the mock request:
image

This screenshot shows a mock response with an alternate return code:
image

### DIFF
--- a/android/plugins/network/src/main/java/com/facebook/flipper/plugins/network/FlipperOkhttpInterceptor.java
+++ b/android/plugins/network/src/main/java/com/facebook/flipper/plugins/network/FlipperOkhttpInterceptor.java
@@ -228,10 +228,12 @@ public class FlipperOkhttpInterceptor
     if (TextUtils.isEmpty(requestUrl) || TextUtils.isEmpty(method)) {
       return null;
     }
+    final int statusCode = route.getInt("status");
 
     final ResponseInfo mockResponse = new ResponseInfo();
     mockResponse.body = data.getBytes();
     mockResponse.statusCode = HttpURLConnection.HTTP_OK;
+    mockResponse.statusCode = statusCode;
     mockResponse.statusReason = "OK";
     if (headersArray != null) {
       final List<NetworkReporter.Header> headers = new ArrayList<>();
--- a/desktop/plugins/network/MockResponseDetails.tsx
+++ b/desktop/plugins/network/MockResponseDetails.tsx
@@ -239,7 +239,7 @@ export function MockResponseDetails({id, route, isDuplicated}: Props) {
   );
   const [nextHeaderId, setNextHeaderId] = useState(0);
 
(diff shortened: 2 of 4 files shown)


bizzguy:
@passy  Thanks for the feedback.  I ran "yarn fix" and updated the PR.

I'm new to JavaScript, so just for my own understanding, did I accidently change the code or are the new style rules what caused the problem?

bizzguy:
@passy 

The build still seems to be failing on lint checks in files that are not in my commits.  Any advice on how to proceed?  Thanks.

passy:
> I'm not a JavaScript expert, so just for my own understanding, did I accidently change the code or are the new style rules what caused the problem?
It looked like your editor did.

passy:
> The build still seems to be failing on lint checks in files that are not in my commits.

Don't worry about those. Sorry, GitHub is a bit aggressive about surfacing those. This looks good to me now. :)

passy:
> The build still seems to be failing on lint checks in files that are not in my commits.

Don't worry about those. Sorry, GitHub is a bit aggressive about surfacing those. This looks good to me now. :)

android/plugins/network/src/main/java/com/facebook/flipper/plugins/network/FlipperOkhttpInterceptor.java
```diff
@@ -228,10 +228,12 @@ private ResponseInfo convertFlipperObjectRouteToResponseInfo(FlipperObject route
     if (TextUtils.isEmpty(requestUrl) || TextUtils.isEmpty(method)) {
       return null;
     }
+    final int statusCode = route.getInt("status");
```
passy, line 231: I suppose there might be mocks that don't yet have a status set? Should those perhaps default to 200?

bizzguy: @passy  I think you are right.  

The Flipper screen allows entry of any string so it could be empty, not a number, or a number that doesn't qualify as an HTTP status code.  So better editing could be done in Flipper.  I was not able to figure out how to do that. (I'm an Android developer and not very skilled in React).

The Android plugin should definitely default to 200 if nothing is entered or if a NumberFormatError is thrown.

I'd like to work on that after the PR is published and see if there are any other improvements.

PR #1447, opened 2020-08-10
Add button to copy highlighted requests to routes (mock requests)

bizzguy:
## Summary

Add feature to Network mocks in the Network Plugin which would allow a user to highlight network requests and copy into new routes (mocks).

For many production apps, network requests can contain many headers (easily 20 or more) and a large amount of data  returned in the response (1000's of characters).  Creating mocks for these manually is time consuming and error prone.

It would be better to make mocks automatically by allowing the user to highlight desired requests and have them automatically copied into new routes, also copying the headers and the response data.

## Changelog

Allow user to create new routes by highlighting existing network requests in the Network plugin

## Test Plan

Tested this change manually by running through the following scenario using the sample Android app:

1). Run a GET request from the Sample app.  Verify that the request/response is correct.  Highlight the request to be copied.

request-screen

2). Go to the Mock dialog by clicking on the "Mock" button

mock-screen

3).  Click on "Copy Highlighted Call" to create a mock Route from the selected request.  Verify that the "data" and "headers" tab panels are correct.

mock-screen-2

mock-screen-3

Close the Dialog

4).  Run the request again from the sample app and verify that a mock request is returned with the correct data.

response-mock-1

response-mock-2

### DIFF
--- a/desktop/plugins/network/ManageMockResponsePanel.tsx
+++ b/desktop/plugins/network/ManageMockResponsePanel.tsx
@@ -18,24 +18,24 @@ import {
   colors,
   Panel,
 } from 'flipper';
-import React, {useContext, useState, useMemo, useEffect} from 'react';
+import React, { useContext, useState, useMemo, useEffect } from 'react';
 
-import {Route, Request, Response} from './types';
+import { Route, Request, Response } from './types';
 
-import {MockResponseDetails} from './MockResponseDetails';
-import {NetworkRouteContext} from './index';
-import {RequestId} from './types';
+import { MockResponseDetails } from './MockResponseDetails';
+import { NetworkRouteContext } from './index';
+import { RequestId } from './types';
 
 type Props = {
-  routes: {[id: string]: Route};
+  routes: { [id: string]: Route };
   highlightedRows: Set<string> | null | undefined;
-  requests: {[id: string]: Request};
-  responses: {[id: string]: Response};
+  requests: { [id: string]: Request };
+  responses: { [id: string]: Response };
 };
 
-const ColumnSizes = {route: 'flex'};
+const ColumnSizes = { route: 'flex' };
 
-const Columns = {route: {value: 'Route', resizable: false}};
+const Columns = { route: { value: 'Route', resizable: false } };
 
 const AddRouteButton = styled(FlexBox)({
   color: colors.blackAlpha50,
@@ -88,8 +88,8 @@ const Icon = styled(Glyph)({
 });
 
 // return ids that have the same pair of requestUrl and method; this will return only the duplicate
-function _duplicateIds(routes: {[id: string]: Route}): Array<RequestId> {
-  const idSet: {[id: string]: {[method: string]: boolean}} = {};
+function _duplicateIds(routes: { [id: string]: Route }): Array<RequestId> {
+  const idSet: { [id: string]: { [method: string]: boolean } } = {};
   return Object.entries(routes).reduce((acc: Array<RequestId>, [id, route]) => {
     if (idSet.hasOwnProperty(route.requestUrl)) {
       if (idSet[route.requestUrl].hasOwnProperty(route.requestMethod)) {
@@ -101,14 +101,14 @@ function _duplicateIds(routes: {[id: string]: Route}): Array<RequestId> {
       };
       return acc;
     } else {
-      idSet[route.requestUrl] = {[route.requestMethod]: true};
+      idSet[route.requestUrl] = { [route.requestMethod]: true };
       return acc;
     }
   }, []);
 }
 
 function _buildRows(
-  routes: {[id: string]: Route},
+  routes: { [id: string]: Route },
   duplicatedIds: Array<string>,
   handleRemoveId: (id: string) => void,
 ) {
@@ -148,12 +148,12 @@ function RouteRow(props: {
           <Icon name="caution-triangle" color={colors.yellow} />
         )}
         {props.text.length === 0 ? (
-          <TextEllipsis style={{color: colors.blackAlpha50}}>
+          <TextEllipsis style={{ color: colors.blackAlpha50 }}>
             untitled
           </TextEllipsis>
         ) : (
-          <TextEllipsis>{props.text}</TextEllipsis>
-        )}
+            <TextEllipsis>{props.text}</TextEllipsis>
+          )}
       </FlexRow>
     </FlexRow>
   );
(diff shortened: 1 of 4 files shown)


passy:
Woah, that's super cool! Thanks for working on this and the detailed test plan!

I've only had a cursory look but want to import this straight away to get feedback. :)

bizzguy:
@passy   Thanks for the feedback.  

I still seem to be having some weird lint problems.  There are lint errors appearing in the build report that don't seem to be problems in my original code.  Mostly related to spaces inside squiggly braces.  Is it possible that something in the build chain is formatting the code incorrectly.  Or is it just something I did wrong.  

I did run "yarn fix" on the "desktop" directory before pushing the PR.

Also, there are some UI changes that still need to be done to "MockResponseDetails".  If there are more than 10 or so headers they go beyond the bottom of the panel instead of scrolling.  I'm planning to work on that next.

desktop/plugins/network/ManageMockResponsePanel.tsx
```diff
@@ -185,7 +198,7 @@ export function ManageMockResponsePanel(props: Props) {
     }
     setCurrentRouteSize(routeSize);
   }, [routes]);
-  const duplicatedIds = useMemo(() => _duplicateIds(routes), [routes]);
+  const duplicatedIds = useMemo(() => _duplicateIds(props.routes), [routes]);
```
passy, line 201: Just one nit here: Shouldn't the dependency here also change from `routes` to `props.routes` to ensure this does update correctly?

bizzguy: I think you're right, I'll make that change.  Thanks.

bizzguy: @passy. I made the fix but it caused a crash in testing so I reversed it.

bizzguy: It appears to be checking (and memorizing) the list of duplicate ids.  Which seems to be working fine.  

"routes" is assigned from "props.routes" in line 189.

bizzguy: @passy Looks like there is a problem with routes created through "copy highlighted".  I'm now starting to get an error message that occurs when I update a route that was copied.  It looks to be in an NDK library on the device (but that's just my guess).  Here's the crash report:

image

bizzguy: @passy   I have not been able to reproduce the error.  For now, I'm assuming it was just a problem for me.

However, while researching empty string problems (which is what the stack trace seems to suggest is the issue) I did find a related edge case.  How should an empty value for an HTTP header be handled?

The Flipper plugin was not adding the header if either the name or value was empty.  But according to the HTTP specification, an empty value is allowed:

[](https://stackoverflow.com/questions/12130910/how-to-interpret-empty-http-accept-header)

bizzguy: @passy. Some additional info on your comment.

You asked about this line:

  const duplicatedIds = useMemo(() => _duplicateIds(props.routes), [routes]);

You  asked if the 2nd occurence of "routes" should also be "props.routes".  The short answer is yes and I did make the change.

After consulting with someone who knows react, they pointed out that by changing the Props object, the deconstruction of routes no longer works the same, so I had to remove this line as well:

    const {routes} = props;

Now everything seems to work properly.

Also, was wondering how you added the code to your comment.  I'd like to be able to do that also.

bizzguy:
@passy I updated this PR to fix the merge conflicts with the body-format PR.  I'm still getting an error in the checks but it appears to be outside of my changes.  Is there anything more I need to do?  Thanks.

passy:
Brilliant. Thanks for making the changes! And really sorry for the delays in the review. It's been a busy few weeks!

desktop/plugins/network/ManageMockResponsePanel.tsx
mweststrate: @bizzguy I might be missing something, but things should behave exactly the same when and when not destructuring `routes`, as `props` is a new object every render. If it does matter, it sounds like there is something side-effectful happening during render? So this render component should behave exactly the same with and without this change, and if there is a change, I'd be a little bit concerned :)

desktop/plugins/network/index.tsx
```diff
@@ -387,6 +404,40 @@ export default class extends FlipperPlugin<State, any, PersistedState> {
...
+              const nextRouteId = draftState.nextRouteId;
+              draftState.routes[nextRouteId.toString()] = {
+                requestUrl: requests[row].url,
+                requestMethod: requests[row].method,
+                responseData: responseData as string,
+                responseHeaders: headers,
+                responseStatus: responses[row].status.toString(),
+              };
+              draftState.nextRouteId = nextRouteId + 1;
+              informClientMockChange(draftState.routes);
```
mweststrate, line 436: This is an awesome feature! Thanks for building it :)

Regarding `informClientMockChange`: This should best be done after the produce, as otherwise the draft is leaked from the produce, which will throw exceptions if any async reference is made the to the routes. So this would become:

```
setState(state => {
   const nextState = produce(state, /* same fn */)
   informClientMockChange(nextState.routes)
   return nextState
})
```

Note that using a side effect in setState is an anti pattern in itself, so instead of the above, the idiomatic thing to do is `useEffect(() => { informClientMockChange(nextState.routes) }, [nextState.routes])`, but that is a higher impact change as multiple code paths lead there, so I can live with the first proposed solution as well

shoeguy: @mweststrate Thanks for the feedback.  I'm an Android developer so struggling a bit with the JavaScript.

I made the changes you suggested (I think!) but having some issues with it.  Here's the code for the function "copyHighlightedCalls" along with the error message.  I hoping you can give me some direction on it.  Thanks

image

mweststrate: should be `produce(state, (state: State) => /* etc */`

shoeguy: @mweststrate You're a guru!  Made the change and tested it.  Just pushed so it should show up shortly.

bizzguy:
@passy  Notice that this PR is marked as "Changes Requested".  I don't see anything I need to do but just wanted to make sure there was nothing.   Thanks

passy (approved):
Sorry! This does look good, I'm just waiting for internal sign-off for this to go in. :)

PR #1479, opened 2020-08-18
[Flipper] Introduced first class console to help users debugging issues

mweststrate:
## Summary

Handling issues typically start with: did you look at the Electron logs? Since Flipper is such an extensible tool, running in varying environments I think the console should be support as first class concept. Many errors are currently not shown to the user. This PR is a first attempt to fix that. 

The implementation is based on https://github.com/samdenty/console-feed, which is used by quite some web based IDE like tools (like codesandbox), and offers a lot of goodies out of the box, like collapsing errors, objects, etc.

Edit: also added a counter keeping track of the amount of errors

N.B. no need to immediately review this diff, I'll import it to phabricator as soon as I can :)

## Changelog

changelog: Introduce 'Debug Logs' section to help users to troubleshoot issues or to provide more accurate reports.

## Test Plan

Screenshot from 2020-08-18 15-29-55


passy:
Yay, Flipper is getting some love on Linux again!

Issue #1476, opened 2020-08-17
Network Plugin does not load mock routes on connection to device


bizzguy:
## 🐛 Bug Report

Network Plugin does not load mock routes on connection to device.  Since the app does not persist the mocks, they are lost when the app is closed.

## To Reproduce

Start the Android Sample app
Create a mock in the Network plugin
Verify that the mock is working using the Android Sample app
Close the Android Sample Flipper app and verify that it no longer appears in Flipper
Execute a network request that has been mocked
The request is not mocked

Note:  The problem appears to be that the mocks (routes) are only updated in the app when they are created or modified.  The app does not persist the mocks.  A fix would be to load the mock table into the app when the plugin connects.

## Environment

Flipper 0.52.1

mweststrate:
Fix will be part of next release, closing issue.

PR #1486, opened 2020-08-21
Network Plugin should load routes into client on init

shoeguy:
## Summary

When the network plugin is loaded, the route table in the device is empty.  It needs to be populated with whatever routes have already been created and saved in local storage.  Otherwise, the user would need to modify the routes to force an update.

Issue is described here - https://github.com/facebook/flipper/issues/1476

## Changelog

Routes updated in device when plugin is initialized

## Test Plan

1. Create a mock for the get request in the sample Android app
2. Kill the app
3. Start the app 
4. Issue the get request in the app
5. Verify that the request is mocked in the Network plugin

image


mweststrate:
Superseded by #1491

PR #1527, opened 2020-09-15
Add chocolatey as a way to install openssl for windows

jknoxville:
We've had several reports that Chocolatey-provided openssl works well with Flipper, so recommending it here.


passy (approved):
Love it!

Issue #1466, opened 2020-08-14
Network Plugin can't parse UTF8 characters and can't decode images properly


skychx:
## 🐛 Bug Report
Network Plugin can't parse UTF8 characters and can't decode images properly.

### 1.Test Link: 
https://ex.ke.com/sdk/recommend/html/100001314?hdicCityId=110000&paramMap[source]=&id=100001314&mediumId=100000037&elementId=&resblockId=1111027381003&templateConfig=%5Bobject%20Object%5D&fbExpoId=346620976471638017&fbQueryId=&required400=true&unique=1111027381003&parentSceneId=

### 2.Test Code(React Native): 

```javascript
fetchData = () => {
  fetch('https://ex.ke.com/sdk/recommend/html/100001314?hdicCityId=110000&paramMap[source]=&id=100001314&mediumId=100000037&elementId=&resblockId=1111027381003&templateConfig=%5Bobject%20Object%5D&fbExpoId=346620976471638017&fbQueryId=&required400=true&unique=1111027381003&parentSceneId=')
}

render() {
  return (
    <TouchableOpacity onPress={this.fetchData}>
      <View></View>
    </TouchableOpacity>
  )
}
```

### 3.Flipper Network Plugin parse result(Error)
<img width="1102" alt="not" src="https://user-images.githubusercontent.com/24872439/90215325-e976e800-de2d-11ea-852d-7099b538dfa0.png">

### 4. Browser parse result(correctly)
<img width="701" alt="yes" src="https://user-images.githubusercontent.com/24872439/90215430-28a53900-de2e-11ea-803e-8b1367dab9b8.png">

### 5.image displays

Version 0.51 decodes and displays images correctly, version 0.52 shows (empty).Sometimes images are decoded into `String` in the absence of `content-type`

> **0.51 ⬇️**
<img width="1133" alt="0 51 0_Image" src="https://user-images.githubusercontent.com/24872439/90215575-95b8ce80-de2e-11ea-8b1a-ecb2978dca0b.png">

> **0.52 ⬇️**
<img width="1120" alt="0 52 0_Image" src="https://user-images.githubusercontent.com/24872439/90215579-981b2880-de2e-11ea-989f-3ddf3a3ead72.png">

> **0.52 image decode into string ⬇️**
<img width="1115" alt="0 52 0_ImageP" src="https://user-images.githubusercontent.com/24872439/90215696-019b3700-de2f-11ea-8768-94e427ad839f.png">

## Environment
[...]

PR #1541, opened 2020-09-22
[NetworkPlugin] Fix body decode issue in Chinese character #1466

yaoandy107:
## Summary
Related to this [issue](https://github.com/facebook/flipper/issues/1466)

## Test Plan
I have written a minimal backend with Python Flask for testing.

Test API code: [https://gist.github.com/yaoandy107/1dc0daf2071aed0448e39100145b1ec3](https://gist.github.com/yaoandy107/1dc0daf2071aed0448e39100145b1ec3)

### DIFF
--- a/desktop/plugins/network/utils.tsx
+++ b/desktop/plugins/network/utils.tsx
@@ -24,7 +24,7 @@ export function decodeBody(container: Request | Response): string {
     return '';
   }
 
-  const b64Decoded = atob(container.data);
+  const b64Decoded = base64DecodeToUnicode(container.data);
   try {
     if (getHeaderValue(container.headers, 'Content-Encoding') === 'gzip') {
       // for gzip, use pako to decompress directly to unicode string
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks for this PR! I'll pick it up further and check if it addresses correctly all the test cases I've been gathering and won't break iOS. Might take me a while, so apologies in advance for that😅

mweststrate:
The fix in this PR is incorrect, as it fix one specific case, but breaks others. But working on more fundamental fix so it is in progress :)

yaoandy107:
@mweststrate Thanks for reviewing, and I'm curious about what situations will break, it look fine in mine daily work. Also, looking foward to see the final fix!

mweststrate:
Should be fixed by https://github.com/facebook/flipper/commit/6b7b1fab5cae3adfdfab89055ec97c9ee8e8fb41 and part of next weeks release. If it isn't solved in Flipper0.64, feel free to open a fresh issue with an example URL

PR #1566, opened 2020-10-01
Fix typo `Abritrary`

vintagesucks:
## Summary

Fixes the typo `Abritrary` in the docs which should be `Arbitrary` instead

## Changelog

Docs: Fixed the typo `AbritraryView` (now `ArbitraryView`)

## Test Plan

--


passy (approved):
Thanks!

PR #1675, opened 2020-11-12
[UPDATE] Catch messages for unconnected plugins

mehul-k:
## Summary

- Maintains a map of all messages to unknown plugin from client side

- Send out all unsentMessages to the plugin when first connected in `initPlugin` in `Client.tsx`

## Changelog

- Use-case where the app was sending message to flipper before the client was connected

- The unsentMessages were flushed and lost forever, now it sends it to the plugin on init for it handle.

## Test Plan

- Send message from app to an unregistered plugin

- `console.log` the `this.unsentMessages` map everytime a new message is added to keep track of the messages

- Register the plugin with flipper and make sure it is receiving all the previous messages, from `this.unsentMessages` map 

`yarn` `yarn start` `yarn test` - all run successfully with no errors reported in flipper console.


mweststrate:
@mehul-k could you elaborate a little bit on "Use-case where the app was sending message to flipper before the client was connected", the premise so far is that messages should not be send to Flipper until you get a connection, and plugins should stop sending messages once they loose a connection. Can you share a little about the plugin in question? What does the code that handles connections or send messages look like?

mehul-k:
@mweststrate the plugin in question sends a `init` message to flipper as soon a connection to flipper is established, but due to the structure of my plugin, there is a delay in registering the plugin. This solves that problem and keeps that initial message until the plugin is detected by flipper instead of dropping it.

mweststrate:
Can you please elaborate on this as well "...but due to the structure of my
plugin, there is a delay in registering the plugin. "

Note that this is quite a fundamental change, so we need to make sure that
A) there is a really good semantic reason for making this change, and B)
once we agree on that, this behavior should be verified with tests
(probably in messagequeue.node.tsx). In principle it should never be the
case that you get a connection _before_ the plugin is initialized in
Flipper, so please do share some more info.

mweststrate:
@mehul-k  any updates on above comments?

mweststrate:
Closing for inactivity.

PR #1735, opened 2020-12-02
Bump OpenSSL-Universal pod dependency to 1.1.171

Iskander508:
## Summary

This should fix the react-native issue with building on Xcode 12: https://github.com/facebook/react-native/issues/29984

## Changelog

Upgrade OpenSSL-Universal dependency on the one that includes arm for iPhone Simulator: https://github.com/krzyzanowskim/OpenSSL/releases

## Test Plan

This PR is not intended to be merged directly, as more version bumping and propagation through the other pods will be needed.

### DIFF
--- a/iOS/Podspecs/Flipper-Folly.podspec
+++ b/iOS/Podspecs/Flipper-Folly.podspec
@@ -16,7 +16,7 @@ Pod::Spec.new do |spec|
   spec.dependency 'boost-for-react-native'
   spec.dependency 'Flipper-Glog'
   spec.dependency 'Flipper-DoubleConversion'
-  spec.dependency 'OpenSSL-Universal', '1.0.2.20'
+  spec.dependency 'OpenSSL-Universal', '1.1.171'
   spec.dependency 'CocoaLibEvent', '~> 1.0'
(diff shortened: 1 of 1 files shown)


passy:
Thanks!

Iskander508:
@passy Thank you for integrating this change, but an updated cocoapod needs to be released. The latest ([2.3.0](https://cocoapods.org/pods/Flipper-Folly)) still points to the older version of OpenSSL-Universal (1.0.2.20)

andreialecu:
I was able to use the updated pod in react-native via:

```rb
  use_flipper!('Flipper-Folly' => { :podspec => 'https://raw.githubusercontent.com/facebook/flipper/master/iOS/Podspecs/Flipper-Folly.podspec' } )
```

If you don't use react-native, something like this should work:
```rb
pod 'Flipper-Folly', :podspec => 'https://raw.githubusercontent.com/facebook/flipper/master/iOS/Podspecs/Flipper-Folly.podspec'
```

andreialecu:
Seems that this fix doesn't take care of the problem entirely. Running into #1758 now - same problem, but a different pod.

PR #1776, opened 2020-12-21
Network plugin - fix issue with responses with empty body

bizzguy:
## Summary

The Network Plugin does not properly handle network requests that return an empty body (because of the body actually being empty or because the network call returns something like a 404 status).

Also, the creation of mocks using the "Copy Highlighted" command when the original response returns an empty body is not handled properly.

## Fix

The Android plugin now returns a response when the body length is 0.

The client plugin creates a body containing an empty string instead of null when the body is empty.

## Changelog

Fix problem in Network Plugin when original call or mock has an empty body in the response.

## Test Plan

The following screen output for the Network Plugin shows that the call is now being handled correctly.  The fields for "status", "size" and "duration" are now populated.

Also, the mock calls are properly defined and shown in yellow.

image

### DIFF
--- a/android/plugins/network/src/main/java/com/facebook/flipper/plugins/network/NetworkFlipperPlugin.java
+++ b/android/plugins/network/src/main/java/com/facebook/flipper/plugins/network/NetworkFlipperPlugin.java
@@ -69,7 +69,7 @@ public class NetworkFlipperPlugin extends BufferingFlipperPlugin implements Netw
             int numChunks =
                 responseInfo.body == null
                     ? 1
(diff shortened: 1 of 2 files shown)


passy:
Happy new year, @bizzguy! Thanks so much for the fix and apologies for the delay. Most of us were off during the holidays. :)

passy (approved):

bizzguy:
@passy  You're welcome!  I did notice that the merge is blocked because of some failing tests but I was not able to see the tests (they are Facebook Internal).  Is there anything I need to do?

passy:
@bizzguy That was just our formatter complaining. :) Should be landing within the next hour or so.

Issue #1674, opened 2020-11-11
iOS network observer has thoroughly broken swizzling


LeoNatan:
## 🐛 Bug Report
The iOS network observer has thoroughly broken swizzling; when it calls the original delegate method, it sends the incorrect selector, causing errors

<img width="1724" alt="Screen Shot 2020-11-11 at 22 25 23" src="https://user-images.githubusercontent.com/2270433/98860867-fd51da80-246c-11eb-87df-c2b34140f827.png">

`_cmd` here is incorrect. It should be `URLSession:dataTask:didReceiveResponse:completionHandler:`, but as you see, it's something else. This is broken behavior, and breaks code that use the well known pattern `[proxyObj respondsToSelector:_cmd]`.

## To Reproduce
Put a breakpoint in any `NSURLSession` delegate method.
Print `_cmd`.

## Environment
iOS 13.5 and iOS 14.2

LeoNatan:
This is most likely happening in other places, where similar careless swizzling behavior has been used.

mweststrate:
@LeoNatan we are not use all iOS versions internally, so when the APIs change we might need to add a rule more to the swizzling. Would you mind creating a PR with what the correct swizzling behavior should be?

LeoNatan:
I will take a look to see if I can come up with a PR. Thanks

arnauddorgans:
Any updates? :)

LeoNatan:
Sorry, I have not had the chance to take a look.

LeoNatan:
I tried looking today, but unfortunately the system is too broken and too (needlessly) complex. I will not be able to provide a fix. The entire swizzling method here is broken. It even breaks calls to `_methodDescription` once `FLEXNetworkObserver` is initialized. Please note that this is blocking Detox and Detox Instruments, and I will just swizzle out `injectIntoAllNSURLConnectionDelegateClasses` as a workaround, meaning users will be unable to use `FLEXNetworkObserver` with Detox.

LeoNatan:
You can easily reproduce by running
```sh
react-native init Bla
```

Then after calling `InitializeFlipper`, add the following line:
```objc
NSLog(@"%@", [self valueForKey:@"_methodDescription"]);
```

You will see that this crashes the system now. If you comment out the `InitializeFlipper` call, it works as expected.

LeoNatan:
[...]

PR #1810, opened 2021-01-05
Fixes a crash caused by network swizzling in FlipperKit

priteshrnandgaonkar:
## Summary

This PR fixes the crash mentioned [here](https://github.com/facebook/flipper/issues/1674#issuecomment-754402609) in the #1674 .  The issue was that logging `NSLog(@"%@", [self valueForKey:@"_methodDescription"]);` in the swizzled network delegate class used to cause a crash. The cause of this was the nil `methodDescription.types` sent to `class_addMethod`.

In the following change we do not add a method to the class which doesn't implement the selector, we will anyway listen for those events from the base NSURLSession class. Removing the `class_addMethod`, fixes both the issue mentioned 
 [here](https://github.com/facebook/flipper/issues/1674#issuecomment-754402609) in #1674.

This PR also fixes an issue where no network calls where logged in Flipper when the iOS version was 14, the reason was that the URLSessionResume task was not properly swizzled for 14.

## Changelog

- Do not add a runtime method to the delegate which doesn't implement a responder
- Fix swizzling logic of the URLSession task resume method for ios 14

## Test Plan

- I ran the sample app and checked that the network calls are logged.
- I copy pasted the changes in RN project and verified that the  issues mentioned in #1674 doesn't happen and also verified that the network calls being hit from JS and Native code gets logged in Flipper. Verified with IsolatedSwizzles.zip mentioned over [here](https://github.com/facebook/flipper/issues/1674#issuecomment-754726174). Also verified by copy pasting the code in the fresh  RN project.


LeoNatan:
I will test tomorrow, thank you!

mweststrate:
@LeoNatan, awesome! Let us know your findings.

priteshrnandgaonkar:
@LeoNatan, did you get the chance to check the changes ? did it work ?

LeoNatan:
@priteshrnandgaonkar Will test now. Had a family medical emergency yesterday. Thanks

LeoNatan:
`_cmd` is still broken:

```
(lldb) po _cmd
"_flex_swizzle_67f1997c_URLSession:dataTask:didReceiveData:"
```

Also, I don't think the behavior is correct. Notice that the FLEX methods are called twice, incorrectly:

```
(lldb) po [NSThread callStackSymbols]
<_NSCallStackArray 0x600003a0eca0>(
0   ???                                 0x00000001116129c7 0x0 + 4586547655,
1   DelegateBreakage                    0x000000010b729490 main + 0,
2   DelegateBreakage                    0x000000010bba7e2f __69+[FLEXNetworkObserver injectTaskDidReceiveResponseIntoDelegateClass:]_block_invoke.354 + 63,
3   DelegateBreakage                    0x000000010bb9dc2c +[FLEXNetworkObserver sniffWithoutDuplicationForObject:selector:sniffingBlock:originalImplementationBlock:] + 444,
4   DelegateBreakage                    0x000000010bba7a04 __69+[FLEXNetworkObserver injectTaskDidReceiveResponseIntoDelegateClass:]_block_invoke_2 + 756,
5   DelegateBreakage                    0x000000010b726dc0 -[__detox_sync_URLSessionDelegateProxy URLSession:dataTask:didReceiveResponse:completionHandler:] + 704,
6   DelegateBreakage                    0x000000010bba7e2f __69+[FLEXNetworkObserver injectTaskDidReceiveResponseIntoDelegateClass:]_block_invoke.354 + 63,
7   DelegateBreakage                    0x000000010bb9dc2c +[FLEXNetworkObserver sniffWithoutDuplicationForObject:selector:sniffingBlock:originalImplementationBlock:] + 444,
8   DelegateBreakage                    0x000000010bba7a04 __69+[FLEXNetworkObserver injectTaskDidReceiveResponseIntoDelegateClass:]_block_invoke_2 + 756,
9   CFNetwork                           0x00007fff236ca9d5 _CFNetworkHTTPConnectionCacheSetLimit + 158957,
10  libdispatch.dylib                   0x000000010f0917ec _dispatch_call_block_and_release + 12,
11  libdispatch.dylib                   0x000000010f0929c8 _dispatch_client_callout + 8,
12  libdispatch.dylib                   0x000000010f099296 _dispatch_lane_serial_drain + 796,
13  libdispatch.dylib                   0x000000010f099f9d _dispatch_lane_invoke + 493,
14  libdispatch.dylib                   0x000000010f0a5de2 _dispatch_workloop_worker_thread + 882,
15  libsystem_pthread.dylib             0x00007fff60c8aa3d _pthread_wqthread + 290,
16  libsystem_pthread.dylib             0x00007fff60c89b77 start_wqthread + 15
)
```

priteshrnandgaonkar:
[We agreed](https://github.com/facebook/flipper/issues/1674#issuecomment-754390731) that _cmd shouldn't be used ? Its a hidden variable, and when you swizzle the selector calling the implementation will the swizzled one.

LeoNatan:
All these issues can be avoided if you adopt a swizzling strategy similar to ours. Create a dynamic subclass that implements all the methods you are interested in, then ISA swizzle the instance to be your dynamic class and calling the super impl if the original object responds to the method.

LeoNatan:
There has been no agreement on breaking the language. You are swizzling incorrectly. Apple themselves used `_cmd` in their own documentation example code.

Normal swizzling does not break the `_cmd`. I can create an example project to show you.

priteshrnandgaonkar:
> There has been no agreement on breaking the language. You are swizzling incorrectly. Apple themselves used _cmd in their own documentation example code.

Show me the written doc stating that the language agreement is broken with the change in _cmd. I disagree with this. _cmd shows the which selector calls the implementation.In the case of swizzling, if one uses exchange_implementation method then _cmd changes and it shows correctly that  the swizzled selector calls the implementation. I suggested the workaround for you to use @selector instead of _cmd.

Also the method followed for swizzling is from the reputed blog of [NSHipster](https://nshipster.com/method-swizzling/).

priteshrnandgaonkar:
The current swizzling code correctly calls back the delegate calls and works fine and I tested with your sent project. What other problem are you facing ?

LeoNatan:
I just tested and it appears you are correct. `_cmd` is different. That's interesting. I already used the workaround in my code. This is the first time I see selectors change with `method_exchangeImplementations`, which is very interesting.

Please test to see that duplicate calls to FLEX methods don't break your internal logic of flipper.

priteshrnandgaonkar:
> Please test to see that duplicate calls to FLEX methods don't break your internal logic of flipper.

I already tested it and it works fine for us. Those calls are to the blocks which are passed from that function. I have checked both in RN and non-RN project and it works fine for us.

LeoNatan:
👍

Final question/request. Since older RNs include broken flipper, I will have to keep my disable code for those versions. I wonder if there is a way, in runtime, to query the version of flipper/FLEX, so that I can disable FLEX only in those older versions? Thanks

priteshrnandgaonkar:
The swizzling code is shipped with FlipperKit. So you can check the version of FlipperKit in your podfile and [disable the networkplugin](https://github.com/facebook/flipper/blob/master/FlipperKit.podspec#L146) instead.  We will make a release with the above change and update it in RN, so that these changes get shipped.

LeoNatan:
Unfortunately, DetoxSync needs to perform this at runtime. It's used in Detox, but also in Detox Instruments, and both projects embed it without any user intervention (to spare our Xcode-challenged users the hassle). I think there is a way to get the RN version in runtime, so I can fall back to that, once it's released. Thanks

mweststrate:
Note that you can also error and point users to here, bumping the SDK is
pretty straight forward and has never introduced issues so far:
https://fbflipper.com/docs/getting-started/react-native#using-the-latest-flipper-sdk

LeoNatan:
Thank you guys both for looking into the problem!

mweststrate:
Landing now, should be rolled out as part of next week's 0.71.0 release

PR #1816, opened 2021-01-07
Move title bar functionality to rail

mweststrate:
Summary:
Now that Sandy is the default in OSS builds as well, we can remove the temporarily title bar and switch to small topbar windows in Electron.

This diff removes any remaining elements in the titlebar

- version number went into the title bar
- the update check warning is shown on top of the bottom section of the left rail (orange triangle)
- the development only perf graphs are moved to the left bar as well and are now aligned vertically.

Differential Revision: D25805957


amyjo7:
Thank you

Issue #1460, opened 2020-08-12
In Network Plugin headers in mock requests are not scrolling


bizzguy:
## 🐛 Bug Report

The Network plugin provides feature to create mock requests containing data and header values.  When multiple header values are entered, they display beyond the end of the enclosing view.  They should become scrollable instead.

Here is a screen shot showing the problem.  The headers extend beyond the end of the enclosing view.  And the "Close" button is also pushed outside the enclosing view.

![image](https://user-images.githubusercontent.com/337874/90058574-84918580-dca7-11ea-8385-446961df57fe.png)

## To Reproduce

Create a mock call and add multiple headers.  Keep adding headers until the header list extends beyond the end of the enclosing view.

## Environment

Flipper Desktop 9.0.2-dev

bizzguy:
I tried fixing this myself.  I thought it would be easy.  The headers are shown in a "ManagedTable" element:

![image](https://user-images.githubusercontent.com/337874/90059838-69277a00-dca9-11ea-9ad6-cd63dabd6403.png)

According to the doc, ManagedTable elements can be made scrollable by just changing the "autoHeight" property:

![image](https://user-images.githubusercontent.com/337874/90059933-89efcf80-dca9-11ea-8045-4fc6c5641d94.png)

But changing the property actually caused all the headers to disappear.

I'd like to continue working on this but don't have much React experience (I'm an Android developer).  Would be interested in any guidance anyone might have.

cekkaewnumchai:
Thank you very much for reporting both the layout of network mock and `ManagedTable`. We appreciate your help. 

@mweststrate might know more about React, but he's off for a week.

bizzguy:
@mweststrate Some additional info:

It appears that the "AutoSizer" which is part of "ManagedTable" is not calculating the height correctly.  It is setting the height to 0 possibly because on of its containers (maybe "Tab") does not specify the height.  This SO issues seems to describe the problem:

    https://stackoverflow.com/questions/38340054/react-virtualized-autosizer-calculates-height-as-0-for-virtualscroll

bizzguy:
This has been fixed with PR https://github.com/facebook/flipper/pull/1855

Issue #1733, opened 2020-12-01
After pressing the "Add Route/Copy Highlighted Calls" the "untitled" route replaces existing routes.


mgoyanes:
## 🐛 Bug Report
After pressing the "Add Route/Copy Highlighted Calls" the "untitled" route replaces existing routes.

## To Reproduce
Have some mocked routes (5) and press the "Add Route" or the "Copy Highlighted Calls". Sometimes, the new "untitled" route replaces existing routes. This behavior is inconsistent but I can reproduce it several times.

## Expected behavior
The new route should be placed at the top of the list without overriding any existing routes.

## Environment
Flipper desktop app version 0.67.0 but I could reproduce the issue at least from version 0.64.0

mweststrate:
cc @bizzguy  would you mind looking into this one?

bizzguy:
I'll take a look.

bizzguy:
@mgoyanes - I was not able to reproduce the error.  I was using the sample Android app to test.  Behavior does sometimes change when using more complex network calls than those in the sample app.  Would you be able to share the actual API calls you were using?

Also, I did find another error that is caused by responses that don't have a body.  (The POST request in the sample app returns a 404 and no body in the response and causes the problem).  I was able to post a fix for this in PR-1776 (https://github.com/facebook/flipper/pull/1776)

I'll try to do more testing on more complex APIs but if you could provide some more detail on how to recreate that error, that would be really helpful.  Thanks.

mgoyanes:
Hello @bizzguy 
Thanks for looking at this.

I will try to create a video with the behavior I'm seeing.


I don't know if I can ask this but here it goes.
This is what I'm missing in the mock network responses:

1. A toggle to enable/disable the mocking of that request. Just like, the response code, have a toggle there to enable or disable that mock.
2. Mock response alias. Add a field where we could add an alias for that mock. That alias would show on the mocks list (left column) thus allowing us to easily see/go to the mock that we want. Normally my mocks start with https://my.server.ulr/something/somethig/endpoint and I only can see the initial part in the left column. I then need to iterate all over the other mocks to find the one that I'm looking for.
3. Make the mocks list column horizontally scrollable.

Thanks.

bizzguy:
@mgoyanes  I like those ideas!  I'm also working on a feature to save the mocks to a file so they could be imported by another developer.  I'd be interested in your ideas on that.

bizzguy:
[...]

Issue #1487, opened 2020-08-21
Mocks should be saved and restored from local storage by app


shoeguy:
## 🐛 Bug Report

When the Network Plugin loads routes (mocks) from local storage, it loads all routes that have been defined for any app.  It should only load the routes for the app that the plugin is in.

The key to local storage should also contain the application name.  It currently does not:

`    const LOCALSTORAGE_MOCK_ROUTE_LIST_KEY = '__NETWORK_CACHED_MOCK_ROUTE_LIST';`

## To Reproduce
1. Add a mock to the Flipper app
2. Run another app (with the Network Plugin and mocks enabled)
3. Add a mock to the new app
4. Stop the Flipper app and restart it
5. In the Network Plugin for Flipper you will see both mocks defined.

## Environment

Flipper 0.52.1

bizzguy:
This can be closed.  It is implemented in this PR https://github.com/facebook/flipper/pull/1871

PR #1831, opened 2021-01-15
Network Plugin - Update UI for routes 

bizzguy:
## Summary

The current UI for managing mock network request definitions (routes) has some issues
- too small for the amount of information it displays
- no formatting for JSON response bodies
- uses the older design system (not ant)
- no separation between commands for creating routes and the list of routes

The following screen images show these problems.

image

image

## Changelog

Enhance UI for Mocks dialog in Network Plugin

## Test Plan

Ran multiple manual tests using the Android sample app.  Here is a screen image (at 1280 x 720) for the enhanced UI.

image


bizzguy:
@mweststrate  I'm also working on the export/import feature for mocks and wanted to make these UI changes as a separate PR.

I hard-coded the dialog height/width but made sure it still fit on a 1280 x 720 screen.  I also added "Layout.Container" which to some of the components (this is more "ant" compliant, right?).

Any suggestions would be appreciated.

bizzguy:
This PR also seems to fix the problem with adding routes described here https://github.com/facebook/flipper/issues/1733

desktop/plugins/network/MockResponseDetails.tsx
```diff
@@ -322,31 +349,11 @@ export function MockResponseDetails({id, route, isDuplicated}: Props) {
               stickyBottom={true}
               autoHeight={true}
               floating={false}
-              // height={300}
```
mweststrate, line 325: Does this fix https://github.com/facebook/flipper/issues/1460? Otherwise Wrapping a `Layout.ScrollContainer` could fix that I'd expect

desktop/plugins/network/ManageMockResponsePanel.tsx
```diff
@@ -248,11 +259,13 @@ export function ManageMockResponsePanel(props: Props) {
       </LeftPanel>
       <RightPanel>
         {selectedId && props.routes.hasOwnProperty(selectedId) && (
+          <Layout.ScrollContainer vertical style={{height: 500}}>
```
mweststrate, line 262: Why did you need a ScrollContainer here? To fix the header pane? In that case I'd move the scrolling inside the tab, so that the tab switcher stays always visible.

desktop/plugins/network/ManageMockResponsePanel.tsx
```diff
@@ -195,7 +199,7 @@ export function ManageMockResponsePanel(props: Props) {
     props.routes,
   ]);
   return (
-    <Container>
+    <Container style={{height: 580}}>
```
mweststrate, line 202: Note that we now expose `<Layout.Left`, so I suspect the locally styled components aren't largely not needed anymore if you do something like:

```
<Layout.Container height={580}> 
    <Layout.Left>
       <Layout.Container width={200}>
          { /* Left content */ }
        </Layout.Container>
        { /*Right content, probably inside container as well */ }
      </Layout.Left>
 </Layout.Container>
 ```


But if you want to keep the changes small, this is fine as well. Note that you can play with the new layout abstractions by using `View > Flipper Design System` to see some examples.

File: desktop/plugins/network/ManageMockResponsePanel.tsx [hunk 4]
```diff
@@ -217,14 +221,21 @@ export function ManageMockResponsePanel(props: Props) {
               props.responses,
             );
           }}>
-          <Glyph
-            name="plus-circle"
-            size={16}
-            variant="outline"
-            color={colors.blackAlpha30}
-          />
+            <Glyph
```
mweststrate, line 224: Looks like prettier wasn't run on this PR. VSCode (and most other IDEs have a plugin for it). If you have trouble running it, let us know, we can run it on our side as well.

mweststrate (commented):
Hey @bizzguy, thanks for looking into this! Left a few comments.

passy:
Sorry, don't have anything useful to say about the code just wanted to swoop in to say that this is awesome! Thanks for doing this. :)

priteshrnandgaonkar:
@bizzguy, does this PR also tackle this [issue](https://github.com/facebook/flipper/issues/1487).

desktop/plugins/network/MockResponseDetails.tsx
bizzguy: Yes, this does resolve https://github.com/facebook/flipper/issues/1460.

desktop/plugins/network/ManageMockResponsePanel.tsx
bizzguy: This ScrollContainer can be removed.  Done.

desktop/plugins/network/ManageMockResponsePanel.tsx
bizzguy: Is "prettier" the same as "yarn fix" (which I did forget to run)?

bizzguy:
@priteshrnandgaonkar   This PR does not solve the problem of saving the routes properly.  But I am working on that along with a few other Route features.

desktop/plugins/network/ManageMockResponsePanel.tsx
bizzguy: @mweststrate   I would like to keep this change small for now.  Once this is approved I'm planning on putting in the PR for import/export of routes.  Then I'd like to revisit the route UI and add a checkbox to routes for enable/disable.

PR #1834, opened 2021-01-16
Add github actions for dependency pod validations

priteshrnandgaonkar:
## Summary

This PR adds github actions to validate the pods, so that the changes which breaks the validation is not made. This PR also downgrades the version update of OpenSSL, as it doesn't validate the Folly podspec. I have [raised](https://github.com/krzyzanowskim/OpenSSL/issues/113) an issue on the repository which manages the openssl dependency.
## Changelog

- Added githubaction for dependent pod validation
- Downgraded the openssl version, as with upgraded version the folly pod didn't validate.
## Test Plan

CI


passy:
@priteshrnandgaonkar Did you check when the OpenSSL pod was last upgraded? I believe this was to mitigate an issue in RN.

.github/workflows/iOS-pod-lint.yml
```diff
@@ -6,21 +6,69 @@ jobs:
...
 
     steps:
-    - uses: actions/checkout@v2
-    - name: Install Dependences
-      run: pod repo update
-      
-    - name: Lint FlipperKit
-      run: pod spec lint FlipperKit.podspec --use-libraries --allow-warnings --verbose --skip-import-validation
-    
+      - uses: actions/checkout@v2
```
passy, line 9: I remember that these are taking quite a long time and using up our resource pool. Do you think it would be safe to only validate these if the files are changed? We could add that as a spec to the "on" event at the top of the file.

passy (changes requested):
RCing for the two questions. :)

.github/workflows/iOS-pod-lint.yml
priteshrnandgaonkar: Thats a good idea. I will do that.

priteshrnandgaonkar:
> @priteshrnandgaonkar Did you check when the OpenSSL pod was last upgraded? I believe this was to mitigate an issue in RN.

@pascal20100 , OpenSSL pod was updated recently. The update actually didn't fix anything. As the Flipper-Folly pod is not automatically released on cocoapods. Its done manually, the reason its not automated is because, there is hardly any change to these pods. I tried to build Folly pod with latest OpenSSL and it failed to validate. Also the mitigation you are suggesting was just a hunch, I don't think he built it and tested it. Even the RN folks have [asked](https://github.com/facebook/react-native/issues/29984#issuecomment-737382396) to put arm64 arch in the  excluded archs as a solution, thats the recommended solution. The reason is that right now OpenSSL's  doesnt build for arm64 simulators. Once that gets fixed, I will upgrade OpenSSL and Folly

passy (approved):

.github/workflows/iOS-pod-lint.yml
passy: These here are still not scoped to pod-related files. Is that intended?

Issue #1651, opened 2020-11-02
Discussion: Create export/import feature in Network Plugin for Routes (Mock API calls)


bizzguy:
Provide feature to allow developer to export definitions of mock requests so they can be shared with another user.

Provide feature to allow developer to import definitions of mock requests created by another user into the Network Plugin.

Provide feature to allow developer to create Routes (mocks) from network data imported from another user.

Use Case
=======

In large enterprises, a separate QA team may be responsible for testing an app.  App behavior is usually depended on Network API calls.  

When the QA team finds a defect, it would be useful to save the Network calls that are involved in the defective behavior.  (This is already provided by the export/import feature - however, these imported calls can't be turned into mocks)

The developer should be able to import the network requests and convert them to mocks (routes).  This would allow the developer to fix the functionality without having to create test data that might only be available to QA.  QA could also then use the mock data to verify that the defect has been fixed.

Current functionality
=====================

Flipper currently has a method for saving Network calls but definitions of mock requests (routes) are not saved.

When a user imports saved Network calls, a new fake device instance is created in Flipper which allows the user to see network calls. Th However, there is no mechanism for creating mocks from these calls and using them in a "live" Flipper session.

cekkaewnumchai:
These features will be quite useful for people who regularly use mocking functionality on network plugin.

We may not have cycle to do this immediately. However, if you are interested in doing this, you can PR so that we can review it.

bizzguy:
@cekkaewnumchai @mweststrate 

I'm starting to work on this PR.  I'm starting out by adding an "Export" button to the Mock dialog.

![image](https://user-images.githubusercontent.com/337874/103023284-4c326b80-4513-11eb-9a19-ed45bb2c36b3.png)

My plan is to use the "startFileExport" util function as the model:

![image](https://user-images.githubusercontent.com/337874/103023506-b814d400-4513-11eb-97a8-b75e6a3dd941.png)

However, I can't seem to figure out how to get the "store" property from the Network plugin like the MenuBar does. 

I'm hoping you can point me in the right direction.  How do I access the "store" so I can pass it to the export function?

bizzguy:
@mweststrate  I'm attempting to create a PR for saving routes separately from the other data (network, logs, etc).  I'm planning to use the same technique and file format.  The menubar saves all the data and I'm trying to use that technique but getting stuck right away.  Can you offer any help on how to get a "dispatch" object from in a plugin?

mweststrate:
[...]

PR #1855, opened 2021-01-25
Network Plugin - New functions to import, export and clear Routes

bizzguy:
## Summary

In the network plugin, add features to import and export routes as described in issue https://github.com/facebook/flipper/issues/1651

Primary use case is that external testers (such as QA teams) would be able to create test data, convert it to mocks and save the mocks to make bug fixes easier for devs.

Here is a screenshot showing location of buttons to perform import/export (and clearing) of mock routes:

image

Here is another screenshot showing export dialog:

image

## Changelog

Network plugin - Add import and export of routes

## Test Plan

Performed manual testing

- create new mocks
- export mocks
- clear mocks
- import mocks
- verify that mocks still work by making GET/POST requests in sample app

Performed various permutations of above manual tests, including restarting Flipper at various points to ensure that test plan still worked.  Also performed visual inspection of exported files to verify correctness.

Would be very interested in learning how to create automated tests for this functionality.

### DIFF
--- a/desktop/plugins/network/MockResponseDialog.tsx
+++ b/desktop/plugins/network/MockResponseDialog.tsx
@@ -54,7 +54,7 @@ export function MockResponseDialog(props: Props) {
           responses={props.responses}
         />
       </Layout.Horizontal>
-      <Layout.Horizontal>
+      <Layout.Horizontal gap>
         <Button
           compact
           padded
--- a/desktop/plugins/network/MockResponseDialog.tsx
+++ b/desktop/plugins/network/MockResponseDialog.tsx
@@ -7,11 +7,14 @@
  * @format
  */
 
-import {FlexColumn, Button, styled, Layout} from 'flipper';
+import {FlexColumn, Button, styled, Layout, Spacer} from 'flipper';
 
 import {ManageMockResponsePanel} from './ManageMockResponsePanel';
 import {Route, Request, Response} from './types';
 import React from 'react';
+
+import {NetworkRouteContext} from './index';
+import {useContext} from 'react';
 
 type Props = {
   routes: {[id: string]: Route};
@@ -39,6 +42,7 @@ const Row = styled(FlexColumn)({
 });
 
 export function MockResponseDialog(props: Props) {
+  const networkRouteManager = useContext(NetworkRouteContext);
   return (
     <Layout.Container pad width={1200}>
       <Title>Mock Network Responses</Title>
@@ -51,6 +55,31 @@ export function MockResponseDialog(props: Props) {
         />
       </Layout.Horizontal>
       <Layout.Horizontal>
+        <Button
+          compact
+          padded
+          onClick={() => {
+            networkRouteManager.importRoutes();
+          }}>
+          Import
+        </Button>
+        <Button
+          compact
+          padded
+          onClick={() => {
+            networkRouteManager.exportRoutes();
+          }}>
+          Export
+        </Button>
+        <Button
+          compact
+          padded
+          onClick={() => {
+            networkRouteManager.clearRoutes();
+          }}>
+          Clear
+        </Button>
+        <Spacer />
         <Button compact padded onClick={props.onHide}>
           Close
         </Button>
--- a/desktop/plugins/network/index.tsx
+++ b/desktop/plugins/network/index.tsx
@@ -45,6 +45,9 @@ import {URL} from 'url';
 import {MockResponseDialog} from './MockResponseDialog';
 import {combineBase64Chunks} from './chunks';
 import {PluginClient, createState, usePlugin, useValue} from 'flipper-plugin';
+import {remote, OpenDialogOptions} from 'electron';
+import fs from 'fs';
+import electron from 'electron';
 
 const LOCALSTORAGE_MOCK_ROUTE_LIST_KEY = '__NETWORK_CACHED_MOCK_ROUTE_LIST';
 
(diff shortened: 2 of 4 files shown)


mgoyanes:
GREAT. This is very handy. Thanks for your work @bizzguy 

Did the following test:

1. Added 3 routes (A, B and C);
2. Exported the routes;
3. Deleted route A;
4. Imported the previously exported routes;

Actual result:
The routes are imported successfully
It now shows 5 routes. Two of them are duplicated (B and C) and show the duplicate route action.

Expected result:
When importing routes, if we already have an exact copy of one of the routes, skip that route import.
In the above scenario, since we already have routes B and C they wouldn't be imported thus wouldn't be duplicated.

desktop/plugins/network/index.tsx
```diff
@@ -385,6 +394,67 @@ export function plugin(client: PluginClient<Events, Methods>) {
...
+            const file = result.filePath;
+            if (!file) {
+              return;
+            }
+            fs.writeFileSync(
+              file,
+              JSON.stringify(routes.get(), null, 2),
+              'utf8',
+            );
+          });
```
mweststrate, line 452: NIT: a message after success would be nice :)

desktop/plugins/network/index.tsx
```diff
@@ -385,6 +394,67 @@ export function plugin(client: PluginClient<Events, Methods>) {
...
+          properties: ['openFile'],
+          filters: [{extensions: ['json'], name: 'Flipper Route Files'}],
+          title: 'Import Routes',
+        };
+        remote.dialog.showOpenDialog(options).then((result) => {
+          const filePaths = result.filePaths;
+          if (filePaths.length > 0) {
+            fs.readFile(filePaths[0], 'utf8', (err, data) => {
+              if (err) {
+                console.error(err);
```
mweststrate, line 408: nit: you could use `import {message} from "antd"; message.error(...)` here as it gives more visibility to the user, see https://ant.design/components/message/,  but this is fine as well.

desktop/plugins/network/index.tsx
```diff
@@ -385,6 +394,67 @@ export function plugin(client: PluginClient<Events, Methods>) {
...
+                return;
+              }
+              const importedRoutes = JSON.parse(data);
+              importedRoutes?.forEach((importedRoute: Route) => {
+                if (importedRoute != null) {
+                  const newNextRouteId = nextRouteId.get();
+                  routes.update((draft) => {
+                    draft[newNextRouteId.toString()] = {
```
mweststrate, line 416: As @mgoyanes noted, probably the preferred behavior is to reuse routes if a route with the same url + method does exist already?

mweststrate (changes requested):
Great work @bizzguy! This looks super useful I left two nits and a request for change for the route reuse. But after that we can land this straight away. Thanks for introducing these features!

bizzguy:
Thanks for the comments everyone.

@mgoyanes  It seems like you were able to run the code before the PR was merged?  I don't know how to do that.  Is there a way to pull the PR and run it?

desktop/plugins/network/index.tsx
bizzguy: Can I use "message()" like you suggest in your other comment or is there another technique?  Thanks.

mweststrate: Yeah, message.info should be great for this :)

mweststrate:
@bizzguy, you can clone the repository (or your fork) and pull / checkout your branch.

desktop/plugins/network/index.tsx
bizzguy: Completed.  Here's the screenshot showing where the error message displays:

image

desktop/plugins/network/index.tsx
bizzguy: Just to confirm, if the same route is imported we could do either of these:

- use the original definition (and ignore the new import) or
- over-ride the original definition

It seems to me that we should over-ride the original definition since the group of imported routes would be needed to test a particular scenario

bizzguy: although, the user could ensure this by "clearing" all the existing routes before doing the import (now I've talked myself into the first approach)

bizzguy: Right now both routes would be shown and the user could pick the one they want by deleting the other - so maybe no change needed

mweststrate: As a user I'd assume the second

desktop/plugins/network/index.tsx
bizzguy: Completed.  Here's the new screenshot:

image

desktop/plugins/network/MockResponseDialog.tsx
```diff
@@ -51,6 +55,31 @@ export function MockResponseDialog(props: Props) {
         />
       </Layout.Horizontal>
       <Layout.Horizontal>
```
mweststrate, line 57: Oh, another nit, could you add `gap` to this `Layout.Horzontal` to give the buttons some spacing?

bizzguy: Completed.

image

mweststrate:
Landed, should be part of tomorrows release. Thanks again @bizzguy!

desktop/plugins/network/index.tsx
bizzguy: @mweststrate   Thanks for merging this PR.  I still intend to implement the change for over-riding any duplicate routes.  Hopefully will have time in the next few days.

mweststrate:
Yeah, tomorrow is our next release, so figured having the first bit is
already super valuable 😊

PR #1871, opened 2021-01-31
Changes to local storage for mocks and body format

bizzguy:
## Summary

Network calls are normally unique to an application as are the associated mocks.  Currently, the mocks for different applications are combined together and saved to the local store.  It would be better if mocks were saved (and retrieved by application).

This PR adds "appId" to the local store name to save mocks separately for each app.  This is not really the ideal technique since different apps could have the same name.  Android apps use packageId rather than app name (which is what appid is) to uniquely identify an app.  However, package id does not seem to be available to the Flipper client so appid is used instead.

This requirement is described in this issue: https://github.com/facebook/flipper/issues/1487

Also, individual developers often have a preference for how they like to view response data (parsed or formatted).  This PR saved the selected format so that the developer does not have to keep selecting it.  Since this preference is not specific to an app, it is not necessary to save the preference for each app.

## Changelog

Network plugin - save mocks by app
Network plugin - save response body format preference to local storage

## Test Plan

Install two apps (with different names)
Create mocks in each app
Restart Flipper
View the mocks for each app and verify that they are unique to the app

### DIFF
--- a/desktop/plugins/network/index.tsx
+++ b/desktop/plugins/network/index.tsx
@@ -50,6 +50,8 @@ import fs from 'fs';
 import electron from 'electron';
 
 const LOCALSTORAGE_MOCK_ROUTE_LIST_KEY = '__NETWORK_CACHED_MOCK_ROUTE_LIST';
+const LOCALSTORAGE_RESPONSE_BODY_FORMAT_KEY =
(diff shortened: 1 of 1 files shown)


passy:
Amazing! Thanks for continuing to make this better. :)

passy (approved):

Issue #1758, opened 2020-12-12
Apple M1 chip: ARM64 simulator build fails / CocoaLibEvent


andreialecu:
## 🐛 Bug Report

The PR #1735 fixes part of the issue with building Flipper for the Arm64 iOS simulator used on M1 Macs.

Even though the fix above is not yet released, I was able to test it via the mechanism described in this post:
https://github.com/facebook/react-native/pull/30543#issuecomment-743753611

However, linking now fails with this error:

```
in .../ios/Pods/CocoaLibEvent/lib/libevent.a(buffer.o), building for iOS Simulator, but linking in object file built for iOS, file '.../ios/Pods/CocoaLibEvent/lib/libevent.a' for architecture arm64
```

See:

https://github.com/facebook/flipper/blob/031cec299b527975cd1e252177372bb33cb0ae18/iOS/Podspecs/Flipper-Folly.podspec#L20
https://github.com/e314521/CocoaLibEvent

That library seems pretty obscure and doesn't seem maintained at all.

## To Reproduce

Use an M1 mac, replace `use_flipper!` with `use_flipper!('Flipper-Folly' => { :podspec => 'https://raw.githubusercontent.com/facebook/flipper/master/iOS/Podspecs/Flipper-Folly.podspec' } )`. Run build in XCode, see above error.

## Environment

MacOS Big Sur 11.0.1
Mac Mini with M1 chip

andreialecu:
I was able to find this script:
https://github.com/e314521/ios-lib/blob/fa1246cbaa0ea664ee285690aedf551ff7753f85/build-libevent.sh

It seems to be what builds `libevent.a` and the other artefacts. 

Forking the library and updating the build script to also build for the `iPhoneSimulator-arm64` target might be an option.

andreialecu:
I spent some time looking into what it would take to update the binaries to be compatible and here are my findings:

I worked on an updated script that will build `libevent` correctly for all targets, including `arm64` iPhoneSimulator: 
https://gist.github.com/andreialecu/2cc00a5e7212bf3b935afd015d693877

However, `lipo` cannot be used to merge both the iOS and simulator `arm64` slice into the same output file:

```
[...]

PR #1892, opened 2021-02-03
Update OpenSSL pod and remove exclude arch setting

priteshrnandgaonkar:
## Summary

This PR updates the OpenSSL pod and thus updates the corresponding Folly and RSocket pods. The new OpenSSL pod has a support of arm64 artifact for iPhomeSimulators.This PR also updates the replaces the libevent pod which didn't work with arm64 arch for iPhoneSimulator.

## Changelog

- Update OpenSSL pod
- Update Folly dep
- Update RSocket dep
- Replace Libevent pod.

## Test Plan

CI


iOS/Podspecs/libevent.podspec
```diff
@@ -0,0 +1,565 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Because libevent is also relied on by Flipper and *with* OpenSSL bindings,
```
alloy, line 6: So I take it this isn’t correct then? Seeing as I don’t see any changes to enable the libevent OpenSSL bindings, or did I overlook them?

alloy (approved):
Nice 👌 

Have you verified if RN and Flipper will indeed share the dependency? Either way, in the future it might be a good idea to start distributing the libevent podspec via the central repository.

iOS/Podspecs/libevent.podspec
priteshrnandgaonkar: Yeah its nothing was required to be changed. I will remove that comment from the podspec file. Also you can now refer this pod from cocoapod direcctly. I have published it in the cocoapod repository.

alloy: > I have published it in the cocoapod repository.

Ah perfect! :pray:

Fl0p:
+1

hsavit1:
Having trouble with this, discussion here https://github.com/facebook/flipper/issues/1758#issuecomment-777691398

PR #1917, opened 2021-02-10
Network Plugin - fix highlighting issues in routes list

bizzguy:
## Summary

Fix problems with row highlighting on mocks route list.  When selecting a row, the highlight would sometimes be on the wrong row.  Also, when adding routes, the routes would sometimes get added in the middle of the list instead of at the end.

These problems were caused by incorrectly referencing route rows in code.  Sometimes index was used and other times the key was used.  Fixed by consistently using key.

This fixes the problem described in issue https://github.com/facebook/flipper/issues/1733

## Changelog

Network Plugin - fix problems with row highlighting on mocks route list

## Test Plan

Manually test multiple scenarios of adding/deleting/selecting route rows.

Test Plan:

Start with no mocks | No rows should be shown | TRUE
Add a single mock | First (and only row should be highlighted) | TRUE
Delete mock | No rows should be shown | TRUE
Add single mockAdd another mock | 2nd row should be highlighted | TRUE
Add another mock | 3rd row should be highlighted | TRUE
Add 2 more mocks and assign number to each row (5 total) | Last row should be highlighted | TRUE
Select 1st row | 1st row should be highlighted | TRUE
Delete 1st row | 1st row should be highlighed | TRUE
Select last row | last row should be highlighted | TRUE
Delete last row | last row should be deleted and new last row should be highlighted | TRUE
Highlight a middle row (row 3) | middle row should be highlighted | TRUE
Delete middle row | middle row should be deleted and last row should be in its place and should be highlighted | TRUE

### DIFF
--- a/desktop/plugins/network/ManageMockResponsePanel.tsx
+++ b/desktop/plugins/network/ManageMockResponsePanel.tsx
@@ -27,6 +27,11 @@ import {RequestId} from './types';
 import {message} from 'antd';
 import {NUX, Layout} from 'flipper-plugin';
 
+import {Modal} from 'antd';
+import {ExclamationCircleOutlined} from '@ant-design/icons';
+
+//const { confirm } = Modal;
+
 type Props = {
   routes: {[id: string]: Route};
   highlightedRows: Set<string> | null | undefined;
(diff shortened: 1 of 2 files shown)


mgoyanes:
Starting to test @bizzguy 
Thanks for this.

mgoyanes:
@bizzguy Until now, I was unable to reproduce the issue I was having.
Will keep testing.

bizzguy:
@mweststrate Is there anything more I need to do for this PR?  Thanks.

desktop/plugins/network/ManageMockResponsePanel.tsx
```diff
@@ -27,6 +27,11 @@ import {RequestId} from './types';
 import {message} from 'antd';
 import {NUX, Layout} from 'flipper-plugin';
 
+import {Modal} from 'antd';
+import {ExclamationCircleOutlined} from '@ant-design/icons';
```
passy, line 31: I think this is unused?

desktop/plugins/network/ManageMockResponsePanel.tsx
```diff
@@ -27,6 +27,11 @@ import {RequestId} from './types';
...
+
+//const { confirm } = Modal;
```
passy, line 33: Could you remove this, please?

passy (changes requested):

desktop/plugins/network/ManageMockResponsePanel.tsx
mweststrate: Addressed it in phabricator

bizzguy: I removed the code (and also combined the imports).  Hopefully, this is not in conflict with phabricator (what is phabricator?).

mgoyanes: At least at some point, it was an internal Facebook tool. I think it allowed to review code, track bugs, etc...

mweststrate: yeah it is kinda our internal version of Github

bizzguy:
@passy   Thanks for the comments.  Also, an article about your build time tracker just popped up in my feeds.  Looks awesome, I'm definitely going to try it out.

passy (approved):
Thanks for making the changes!

PR #1969, opened 2021-02-26
Network Plugin - Disable routes

bizzguy:
## Summary

Mock routes may contain quite a bit of data for header values and response bodies.  Currently, to turn off a mock a user must delete the mock and then go through the process of creating it when needed again.  This change will allow a user to temporarily disable a mock without deleting it.

A checkbox on the Routes list is used to enable/disable mocks.  As shown in this screenshot:

image

## Changelog

Network Plugin - disable/enable mock routes

## Test Plan

Create mocks using sample app
Verify that mocks are working as expected
Disable the mocks
Verify that the client has been updated with mocks (minus the disabled ones) user Flipper Messages plugin
Use the sample app to send the disabled requests again and verify that they are not mocked

### DIFF
--- a/desktop/plugins/network/ManageMockResponsePanel.tsx
+++ b/desktop/plugins/network/ManageMockResponsePanel.tsx
@@ -102,9 +102,12 @@ function RouteRow(props: {
   handleEnableId: () => void;
   enabled: boolean;
 }) {
+  const tip = props.enabled
+    ? 'Un-check to disable mock route'
+    : 'Check to enable mock route';
   return (
     <Layout.Horizontal gap>
-      <Tooltip title="Check to enable mock route" mouseEnterDelay={1.1}>
+      <Tooltip title={tip} mouseEnterDelay={1.1}>
         <Checkbox
           onClick={props.handleEnableId}
           checked={props.enabled}></Checkbox>
@@ -269,7 +272,6 @@ export function ManageMockResponsePanel(props: Props) {
                   });
                 },
                 (id) => {
-                  console.log('...enableRoute');
                   networkRouteManager.enableRoute(id, {});
                 },
               )}
--- a/desktop/plugins/network/ManageMockResponsePanel.tsx
+++ b/desktop/plugins/network/ManageMockResponsePanel.tsx
@@ -17,20 +17,12 @@ import {
   Panel,
 } from 'flipper';
 import React, {useContext, useState, useMemo, useEffect} from 'react';
-
 import {Route, Request, Response} from './types';
-
 import {MockResponseDetails} from './MockResponseDetails';
 import {NetworkRouteContext} from './index';
 import {RequestId} from './types';
-
-import {message, Modal} from 'antd';
+import {message, Checkbox, Modal, Tooltip} from 'antd';
 import {NUX, Layout} from 'flipper-plugin';
-
-import {Modal} from 'antd';
-import {ExclamationCircleOutlined} from '@ant-design/icons';
-
-//const { confirm } = Modal;
 
 type Props = {
   routes: {[id: string]: Route};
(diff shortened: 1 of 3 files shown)


mgoyanes:
@bizzguy 
Thanks for this. It's really helpful.

Tested and it is working as expected.

Found a minor bug.
image

If the route is selected, hovering on the checkbox shows an incorrect message. It should say "Uncheck to disable mock route".


I also have another suggestion. A button to disable/enable all routes at once. It may come in handy when we have a lot of routes.

desktop/plugins/network/index.tsx
```diff
@@ -370,6 +373,19 @@ export function plugin(client: PluginClient<Events, Methods>) {
         }
         informClientMockChange(routes.get());
       },
+      enableRoute(id: string, routeChange: Partial<Route>) {
+        console.log(routeChange);
```
mweststrate, line 377: could you drop the `console.log` statements?

desktop/plugins/network/index.tsx
```diff
@@ -370,6 +373,19 @@ export function plugin(client: PluginClient<Events, Methods>) {
...
+        console.log(routes.get()[id]);
+        routeChange.enabled = !routes.get()[id].enabled;
+        console.log(routeChange);
+        if (routes.get().hasOwnProperty(id)) {
```
mweststrate, line 381: NIT: `routes.get()[id]` should suffice

desktop/plugins/network/ManageMockResponsePanel.tsx
```diff
@@ -248,19 +253,26 @@ export function ManageMockResponsePanel(props: Props) {
...
+                    onOk() {
+                      const nextId = getNextId(id);
+                      networkRouteManager.removeRoute(id);
+                      setSelectedId(nextId);
+                    },
+                    onCancel() {},
+                  });
+                },
+                (id) => {
+                  console.log('...enableRoute');
```
mweststrate, line 272: Please drop this one :)

desktop/plugins/network/index.tsx
```diff
@@ -370,6 +373,19 @@ export function plugin(client: PluginClient<Events, Methods>) {
         }
         informClientMockChange(routes.get());
       },
+      enableRoute(id: string, routeChange: Partial<Route>) {
```
mweststrate, line 376: I think you can drop the whole `routeChange` argument, and simplify the entire update to (I found the current implementation a bit hard to follow with the empty object that swaps it's enabled attribute):

```
routes.update((draft) => {
   draft[id].enabled = !draft[id].enabled
});
```

mweststrate (changes requested):
I think this is an awesome feature from usability perspective! Thanks for thinking about it. Left a few comments

bizzguy:
> @bizzguy
> Thanks for this. It's really helpful.
> 
> Tested and it is working as expected.
> 
> Found a minor bug.
> image
> 
> If the route is selected, hovering on the checkbox shows an incorrect message. It should say "Uncheck to disable mock route".
> 
> I also have another suggestion. A button to disable/enable all routes at once. It may come in handy when we have a lot of routes.

I made the changes to the tool tip messages but did not add a disable/enable button.  I'm not sure how the UI for that should look so I'd like to think about it a bit more and create a PR for it later.

desktop/plugins/network/index.tsx
bizzguy: I made the change to remove Partial.  However, I'm finding some situations where there is no "id" so I'd like to keep the other code for now (especially since it appears in some of the other existing methods).

bizzguy: The other code being:

`if (routes.get().hasOwnProperty(id)) {`

mweststrate (approved):

Issue #1509, opened 2020-09-01
Listening for react native bundler on a different port


anirudhagarwal365:
I was trying to integrate react native Android application with Flipper. I don't see React Native device on Flipper desktop application. 
I am suspecting this is happening because my metro bundler runs on custom port 8082. How do I configure Flipper to listen on my custom port.

I tried reading up on https://fbflipper.com/docs/custom-ports, but need help understanding INSECURE_PORT & SECURE_PORT.

nikoant:
Hi @anirudhagarwal365, the doc you mentioned describes ports for communication between Flipper desktop app and devices. Metro bundler port is currently [hardcoded to 8081](https://github.com/facebook/flipper/blob/master/desktop/app/src/dispatcher/metroDevice.tsx#L18). I think the easiest solution to fix this is to change this code a bit to make it possible to specify port by env var, e.g. FLIPPER_METRO_PORT.

anirudhagarwal365:
Hi @nikoant. Thanks for pointing it out. Was able to run successfully by creating a fork of the project and making the required changes. Now I am able to see the react native device. (Might create a PR later if I get time)

However, I am not getting "Hermes Debugger Error" - "Metro disconnected. Please check that metro is running and Flipper can connect to it." on the Hermes debugger tab. Logs and React DevTools seem to be working as expected. 

I have already verified that global.HermesInternal !== null, return true.

mweststrate:
Looks like that should be using the env var:
https://github.com/facebook/flipper/blob/master/desktop/plugins/hermesdebuggerrn/index.tsx#L26

anirudhagarwal365:
Thanks @mweststrate. That solves this issue. 

Somehow stuck with "Metro is connected but no Hermes apps were found". 

I tried "http://localhost:8082/json" (8082, since metro server runs on that port), but it returns []. Could this be the issue?

nikoant:
@anirudhagarwal365 do you mind to open PR with your changes?

anirudhagarwal365:
Hi @nikoant . Still not able to successfully run hermes debugger. Still stuck on "Metro is connected but no Hermes apps were found". Not sure if I have accounted for all the changes.

anirudhagarwal365:
Hi @nikoant Can you help here or tag someone who might be able to?

nikoant:
[...]

PR #2050, opened 2021-03-12
Support custom metro ports

matgardon:
## Summary

Based on issue #1509 I needed a custom port for the metro server to pass to flipper.

## Changelog

introduce `METRO_SERVER_PORT` env variable to be able to customize default 8081 port on startup.

## Test Plan

Tested with a locally built linux desktop app, this seems to work OK pour hermes debugging & RN logging. 

React DevTools don't seem to find the running app, maybe there's a mapping to handle there too ?

### DIFF
--- a/desktop/plugins/hermesdebuggerrn/index.tsx
+++ b/desktop/plugins/hermesdebuggerrn/index.tsx
@@ -23,7 +23,10 @@ import ErrorScreen from './ErrorScreen';
 import ChromeDevTools from './ChromeDevTools';
 
 const POLL_SECS = 5 * 1000;
-const METRO_HOST = 'http://localhost:8081';
+const METRO_PORT_ENV_VAR = process.env.METRO_SERVER_PORT || '8081';
+const METRO_PORT = isNaN(+METRO_PORT_ENV_VAR) ? '8081' : METRO_PORT_ENV_VAR;
+const METRO_URL = new URL('http://localhost');
+METRO_URL.port = METRO_PORT;
 
 export type Target = Readonly<{
   id: string;
@@ -87,7 +90,7 @@ export default class extends FlipperDevicePlugin<State, any, any> {
   }
 
   checkDebugTargets = () => {
-    fetch(`${METRO_HOST}/json`)
+    fetch(`${METRO_URL.toString()}json`)
       .then((res) => res.json())
       .then(
         (result) => {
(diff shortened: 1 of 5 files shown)


desktop/plugins/hermesdebuggerrn/index.tsx
```diff
@@ -21,9 +21,12 @@ import Banner, {isBannerEnabled} from './Banner';
 import SelectScreen from './SelectScreen';
 import ErrorScreen from './ErrorScreen';
 import ChromeDevTools from './ChromeDevTools';
+import {parseEnvironmentVariableAsNumber} from '../../app/src/utils/environmentVariables';
 
 const POLL_SECS = 5 * 1000;
-const METRO_HOST = 'http://localhost:8081';
+const METRO_HOST = 'localhost';
+const METRO_PORT = parseEnvironmentVariableAsNumber('METRO_SERVER_PORT', 8081);
+const METRO_URL = `http://${METRO_HOST}:${METRO_PORT}`;
```
jknoxville, line 29: Just to be on the safe side, could you replace this with the following just to make sure no unexpected injections are possible?

```js
const METRO_URL = new URL('http://localhost');
METRO_URL.port = METRO_PORT;
```

and then use `METRO_URL.toString()` to get the url below.

jknoxville (changes requested):
This looks great to me, I just have one request below.

desktop/plugins/hermesdebuggerrn/index.tsx
matgardon: 6fd4b56f3069e5d93f6f145425509cb18918ac5d
I had to cancel importing `parseEnvironmentVariableAsNumber` as I didn't notice it was in different packages.
Using it as a string anyway for `URL` ctor so parsing routine is a bit different.

matgardon: FYI I just figured why I couldn't built locally : I was on node 10 or 12, and metro couldn't build, finally worked it out by switching to node 15. The doc seems outdated on this one (node >= 8).

matgardon:
rebased to up-to-date master from upstream (+autosquash fixups)

matgardon:
Rebased to updated master.
I don't really know if this is useful though and it's taking me more time to keep this PR ready for review/merge than it took to look into the codebase and add the necessary code.
@jknoxville is there a chance this PR is going to be reviewed again in the near future or should I leave it as is and when someone has the time, it will then be useful to update it to current codebase ?

jknoxville:
@matgardon Really sorry for the hold up, this slipped off my radar. Thanks for making the changes, I think we can merge it now.

PR #2080, opened 2021-03-18
Add protobuf support to network inspector

hbmartin:
## Summary

Protobuf based APIs are becoming more common (i.e. gRPC) but are difficult to inspect. Unlike plain text data formats (JSON), Protobuf calls transmit binary data requiring the format to be known ahead of time, making ad-hoc inspection impossible. This PR allows for those format definitions (messages in protobuf terminology) to be transmitted from the client to the network inspector plugin. These definitions are then imported into ProtobufJS which enables the binary data transmitted to be inspected as easily as JSON data.

See Retrofit PR in https://github.com/facebook/flipper/pull/2084

## Changelog

* Add ProtobufJS library to network plugin
* New `ProtobufFormatter` UI in `RequestDetails`
* `ProtobufDefinitionsRepository` to cache and load protobuf defintions
* `addProtobufDefinitions` call in the Android network plugin

## Test Plan

screenshot


passy:
This looks great! Thanks so much for submitting this. I won't have time to review this today but I'll try to take a closer look in the next days. Sorry for the delay! Same goes for your second PR. :)

hbmartin:
Thanks @passy ! Look forward to the feedback 😄

hbmartin:
@passy CI is failing with `error Your lockfile needs to be updated, but yarn was run with '--frozen-lockfile'.`,, I'm not sure how to address this?

desktop/plugins/network/RequestDetails.tsx
```diff
@@ -804,6 +805,85 @@ class BinaryFormatter {
...
+        return (
+          <Text>Could not locate response body data for {request.url}</Text>
+        );
+      }
+    }
+    return undefined;
+  }
+
+  _base64ToArrayBuffer(base64: string): Uint8Array {
+    const binary_string = window.atob(base64);
```
mweststrate, line 877: `window.atob` is typically not what you can use if the base64 encodes binary data rather than an UTF-8 string, and it can crash Flipper pretty hard. Instead, you probably need [`Base64.toUint8Array`](https://www.npmjs.com/package/js-base64#synopsis) (we use the same package in the rest of the network plugin as well)

passy:
@hbmartin Sorry, because this has been taking me a while to revisit there are quite a few conflicts now. Do you think you could take a look? No worries if not, I can do that but it might take a little.

hbmartin:
@passy merge in master, looks like the conflicts were just yarn lock and the move to `public` directory

PR #2106, opened 2021-03-25
ANDROID_SDK_ROOT support

sikrinick:
## Summary
According to Google, ANDROID_HOME is deprecated.
This means, that having only ANDROID_SDK_ROOT as environment variable leading to Android SDK is enough.
Flipper supports only ANDROID_HOME, so I would like to add ANDROID_SDK_ROOT support.
However, there is a set of rules of ANDROID_HOME and ANDROID_SDK_ROOT priority.
Everything is mentioned on a this page: https://developer.android.com/studio/command-line/variables

## Changelog
Added ANDROID_SDK_ROOT environment variable support

## Test Plan
Probably I would need a help with this


passy:
Thanks! That's a great catch.

PR #2116, opened 2021-03-29
App icon for Linux

### LINKED ISSUE #1369: No icon displayed on Ubuntu 20.04
## 🐛 Bug Report
No icon displayed on my Ubuntu 20.04 panel.

## To Reproduce
- Download v0.49.0 for Linux
- Unzip and run
- No icon is displayed on the panel (see screenshot).

flipper_no_icon

## Environment
Flipper 0.49.0, Ubuntu 20.04 (updated from 18.04 LTS).
Previously I used older version of Flipper (0.28.0), then I've deleted the unzipped folder, downloaded a new release and run it.

iqqmuT:
## Summary

Fixes #1369 
It seems that electron-builder has an issue when trying to set the app icon for Linux zip builds. A workaround for this is to set `icon` option for `BrowserWindow`.

I created a new PNG file for this, it seems that 64x64 looks the best.

## Changelog

App icon for Linux fixed.

## Test Plan

Screenshot from 2021-03-29 13-50-48


desktop/static/main.ts
```diff
@@ -279,6 +280,12 @@ function createWindow() {
     minWidth: 800,
     minHeight: 600,
     center: true,
+    // The app icon is defined in package.json by default.
+    // When building Linux zip, it must be defined here or else it won't work.
+    icon:
+      os.platform() === 'linux'
+        ? path.join(__dirname, 'icons/app_64x64.png')
```
passy, line 287: Have you checked if this also works for both development and release builds? There are times where you need to resolve the static path differently because of the ASAR compression.

passy (changes requested):
This looks great. Thanks for looking into this. Just bouncing it back real quick for a question about the path resolution.

desktop/static/main.ts
iqqmuT: I'm not really sure how to know if I'm running a development or release build. However, the icon works in both ways:
* `yarn start`
* `yarn build --linux --version 0.82.1`

passy: Thanks! Just wanted to check that you tested both cases. :)

Issue #2078, opened 2021-03-18
SharedPreferences plugin will crash when key is null


bruce3x:
## 🐛 Bug Report

SharedPreferences plugin will crash when null key fields in shared preference on Android.

## To Reproduce

1. Create a shared preferences file with null key field.

```kotlin
 context.getSharedPreferences("test_sp", Context.MODE_PRIVATE).edit().putInt(null, 123).commit()
```

2. Open Flipper desktop app, load shared preferences plugin, crashed.
```
Name
java.lang.RuntimeException: org.json.JSONException: Names must be non-null
Reason
java.lang.RuntimeException: org.json.JSONException: Names must be non-null
Stacktrace
# 0: 0x79e0d62048 _ZN8facebook7flipper13FlipperClient9callstackEv
# 1: 0x79e0d612f4 _ZN8facebook7flipper13FlipperClient17onMessageReceivedERKN5folly7dynamicENSt6__ndk110unique_ptrINS0_16FlipperResponderENS6_14default_deleteIS8_EEEE
# 2: 0x79e0d70620 _ZN8facebook7flipper28FlipperConnectionManagerImpl17onMessageReceivedERKN5folly7dynamicENSt6__ndk110unique_ptrINS0_16FlipperResponderENS6_14default_deleteIS8_EEEE
# 3: 0x79e0d83638
# 4: 0x79e0d84248
# 5: 0x79e0d9c92c _ZN7rsocket23RSocketResponderAdapter21handleRequestResponseENS_7PayloadEjNSt6__ndk110shared_ptrIN5yarpl6single14SingleObserverIS1_EEEE
# 6: 0x79e0dcca5c _ZN7rsocket19RSocketStateMachine16onNewStreamReadyEjNS_10StreamTypeENS_7PayloadENSt6__ndk110shared_ptrIN5yarpl6single14SingleObserverIS2_EEEE
# 7: 0x79e0dd7258 _ZN7rsocket22StreamStateMachineBase16onNewStreamReadyENS_10StreamTypeENS_7PayloadENSt6__ndk110shared_ptrIN5yarpl6single14SingleObserverIS2_EEEE
# 8: 0x79e0dd500c _ZN7rsocket24RequestResponseResponder13handlePayloadEONS_7PayloadEbbb
# 9: 0x79e0dcbf60 _ZN7rsocket19RSocketStateMachine22onRequestResponseFrameEjNS_7PayloadEb
[...]

PR #2146, opened 2021-03-31
Ignore null key in shared preferences

bruce3x:
## Summary

SharedPreferences plugin will crash when null key fields in shared preference on Android. Null keys should be ignored.

Fixes https://github.com/facebook/flipper/issues/2078

## Changelog

- Ignore null keys in shared preferences.


passy:
Thanks for taking care of this! :)

PR #2152, opened 2021-04-03
Fix for idb file pull command

iqqmuT:
## Summary

Fixes following error when using Flipper with iOS device:

Connection to *App* on *Device* failed.
Failed to start client connection
```
Error: EISDIR: illegal operation on a directory, read
errno: -21
code: "EISDIR"
syscall: "read"
```

The problem is that `idb file pull --bundle-id com.foo.bar src.txt dest.txt` does not create `dest.txt` but `dest.txt/src.txt`. The last argument must be a **directory**, not a file name.

## Changelog

Fixed a bug where Flipper failed to connect the app on iOS device.

## Test Plan

* build idb from sources
* use Flipper with iOS device

### DIFF
--- a/desktop/app/src/utils/CertificateProvider.tsx
+++ b/desktop/app/src/utils/CertificateProvider.tsx
@@ -496,7 +496,7 @@ export default class CertificateProvider {
             deviceId,
             originalFile,
             bundleId,
-            path.join(dir, csrFileName),
+            dir,
             this.store.getState().settingsState.idbPath,
           )
           .then(() => dir);


mweststrate:
Thanks for finding this @iqqmuT! I'm not sure why this didn't cause problems for us, might the internal idb version deviate here @jknoxville?

jknoxville:
Yeah @mweststrate I have seen deviations in the past, so it's possible. That was my thought too, it works well internally so we'll have to test this thoroughly, but this is da very helpful find!

Thanks very much for the submission @iqqmuT !

mweststrate:
Thanks for the PR! During testing we discovered that in not all scenarios this works as expected, and for a proper fix it seems that idb should be re-released first, so we're chasing that but it might take a while. Thanks again and we'll post updates when we have any.

mweststrate:
@iqqmuT a new version of idb got released and of last week! Could you verify if this change is still needed with idb 1.2.2? I wouldn't expect so, but if that can be verified that would be great as our internal version of idb has its own distribution mechanism

iqqmuT:
Issue facebook/idb#659 is still valid for idb v1.1.2, preventing the command `idb file pull --bundle-id` working properly. So let's wait until the issue has been fixed?

mweststrate:
Seems this changed has landed anyway in https://github.com/facebook/flipper/commit/89d7e17c260220e4ac96e03583388af8d652e6b6.

PR #2242, opened 2021-04-27
Added a new bus icon to icons.json

SaqibJDev:
## Summary

For the custom plugin development, I wanted to use the icon represented by a bus as per the plugin's nature. I'm able to see the icon in the flipper but it suggested me to add it in the static/icons.json

## Changelog

New Bus icon for custom plugins

## Test Plan
image


passy:
Thanks for raising this. This is something we need to revisit because our current model for icons expects them all to be part of the centralised bundle to be preloaded. That doesn't really work with decentralised plugins. We're planning to add an option to either use an icon font or an Ant.design icon for this, so you won't need to update the central file anymore.

SaqibJDev:
Hi @passy, thanks for the update. do you if we should expect it in the upcoming release of the Flipper already? Or is there any timeline?

PR #2337, opened 2021-05-18
Detect JSON API content type ('application/vnd.api+json') as textual content

denis-colliot:
## Summary

Network requests and responses using JSON API content type of pattern `application/vnd.api+json` (see https://jsonapi.org/#mime-types) are currently treated as binary data. They should be treated as textual content.

## Changelog

Add regular expression to detect JSON content types.

## Test Plan

Using an API that return `Content-Type` of pattern `application/vnd.api+json`, verify that the response data in sidebar appears as text rather than `(binary data)`.


mweststrate:
Thanks!

PR #2368, opened 2021-05-25
docs: update network-plugin.mdx android

f4z3k4s:
Flipper network plugin does not work without the above added lines.

## Summary

Without the added lines, the network plugin simply does not work on Android with Flipper version 0.91.0. I found this solution in your example react-native project. This should be stated in the docs.

## Changelog

This should not be stated in changelog.

## Test Plan

There is no code.


nikoant:
Thanks for updating docs, @f4z3k4s!

PR #2375, opened 2021-05-27
Fang/master

FFFang:
avoid SuperClass being hooked & code format

## Summary

The UserDefault plugin hook NSUserDefaults's forwardInvocation method,which NSUserDefaults does not implement.It will override root class NSObject's forwardInvocation,which would cause some problems.This behavior would cover some crash es when a class/instance calls a delegate method and not implement the delegate method.

## Changelog

add forwardInvocation method before hook

## Test Plan

No Test Plan


passy:
@priteshrnandgaonkar Could you take a look at this?

aigoncharov:
@lblasa would you have a moment to look at this one?

PR #2377, opened 2021-05-27
Fix search by passing query to search request

arthuralee:
## Summary

When publishing a new flipper plugin, I realized my plugin wasn't showing up in the Plugin Installer search. This was due to a bug where we are only grabbing 50 plugins from NPM and filtering through them locally, instead of passing the search term to the search query. There are 59 plugins that match the keyword filter on npm so 9 plugins never get surfaced.

## Changelog

Fix missing plugins in plugin manager's NPM search

## Test Plan

Test the search feature in Plugin Manager

Search for the previously missing package:
image

Previously working searches continue to work:
image


passy:
Thanks for the fix!

PR #2446, opened 2021-06-14
refactor: configuration on react-native projects

b3coded:
## Summary

This should help to make Flipper can be configured with some default plugins on React Native, removing configurations from the main `react-native` repo.

This takes advantage of the `react-native-flipper` dependency to make possible more easy way to install, remove and update Flipper without needing to change the react-native repo.
This changes may be a Breaking Change.

I'm not familiar with Obj-C, I ran this code on Android Device and iOS simulator and it worked.

## Changelog

- bump Flipper to 0.93 in example app
- added `initializeFlipper` with default plugins to `react-native-flipper`
- added `use_flipper` and `flipper_post_install` to custom ruby script (they should be required in project `Podfile` [example](https://github.com/b3coded/flipper/blob/refactor/initialize-flipper/react-native/ReactNativeFlipperExample/ios/Podfile#L3) (extracted from `scripts/react_native_pods.rb`from ain React Native repo)
- removed Java files from debug variant on Android project
- replaced current `initializeFlipper` with new `initializeFlipper` from `react-native-flipper` on Android and iOS


react-native/ReactNativeFlipperExample/package.json
```diff
@@ -37,4 +37,4 @@
     "browserslist": "^4.16.5",
     "ws": "7.4.6"
   }
-}
+}
```
passy, line 40: Could you revert this change, please?

react-native/react-native-flipper/android/build.gradle
```diff
@@ -84,6 +84,14 @@ dependencies {
     debugImplementation("com.facebook.flipper:flipper:${FLIPPER_VERSION}") {
         exclude group:'com.facebook.fbjni'
     }
+
+    debugImplementation("com.facebook.flipper:flipper-network-plugin:${FLIPPER_VERSION}") {
```
passy, line 88: So I guess the trade-off here is that we make the setup a little simpler while forcing some more deps onto people that they may not want?

react-native/react-native-flipper/android/src/debug/java/com/facebook/flipper/reactnative/ReactNativeFlipper.java
```diff
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-package com.reactnativeflipperexample;
```
passy, line 8: I have a feeling that we had that for some reason. Did it build fine for you after changing this?

passy (commented):
Thanks for the changes. I want to get some more eyes on this to make sure we're not missing anything.

react-native/react-native-flipper/android/src/debug/java/com/facebook/flipper/reactnative/ReactNativeFlipper.java
b3coded: Yes, I ran on my phone and it worked. I will test later on Android Emulator to make sure if its working properly

it's being used [here](https://github.com/facebook/flipper/blob/24527300d42e55cbd7716814c99f5cbaad20d181/react-native/ReactNativeFlipperExample/android/app/src/main/java/com/reactnativeflipperexample/MainApplication.java#L72)

So I changed this file to use `initializeFlipper` from `react-native-flipper` added in this PR.
But the developer can use a customized initialization method, just need to create a custom Java class (like the current ReactNative template does)

react-native/react-native-flipper/android/build.gradle
b3coded: I think we can add some properties to disable some plugins.
So, we add all plugins by default and we can use properties to disable some plugins, example: `FlipperDisablePlugin_Network`
and then use in Java [like this](https://github.com/react-native-async-storage/async-storage/blob/master/android/src/main/java/com/reactnativecommunity/asyncstorage/AsyncStoragePackage.java#L27)

react-native/ReactNativeFlipperExample/package.json
b3coded: Sure, reverted

react-native/ReactNativeFlipperExample/ios/ReactNativeFlipperExample.xcodeproj/project.pbxproj
```diff
@@ -836,6 +836,7 @@
 				COPY_PHASE_STRIP = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				ENABLE_TESTABILITY = YES;
+				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "arm64 i386";
```
priteshrnandgaonkar, line 839: Why are we excluding this archs ?

react-native/ReactNativeFlipperExample/ios/ReactNativeFlipperExample/AppDelegate.m
```diff
@@ -12,35 +12,17 @@
...
 
 #ifdef FB_SONARKIT_ENABLED
-#import <FlipperKit/FlipperClient.h>
-#import <FlipperKitLayoutPlugin/FlipperKitLayoutPlugin.h>
-#import <FlipperKitNetworkPlugin/FlipperKitNetworkPlugin.h>
-#import <FlipperKitReactPlugin/FlipperKitReactPlugin.h>
-#import <FlipperKitUserDefaultsPlugin/FKUserDefaultsPlugin.h>
-#import <SKIOSNetworkPlugin/SKIOSNetworkAdapter.h>
-
-static void InitializeFlipper(UIApplication* application) {
```
priteshrnandgaonkar, line 22: Why this is removed ? I will prefer a c function over a class initialization and a method call.

react-native/ReactNativeFlipperExample/ios/ReactNativeFlipperExample/AppDelegate.m
```diff
@@ -12,35 +12,17 @@
...
 
 @implementation AppDelegate
 
 - (BOOL)application:(UIApplication*)application
     didFinishLaunchingWithOptions:(NSDictionary*)launchOptions {
-#ifdef FB_SONARKIT_ENABLED
-  InitializeFlipper(application);
-#endif
+  
+  #ifdef FB_SONARKIT_ENABLED
```
priteshrnandgaonkar, line 23: You should ideally add one more check  of `DEBUG` as your class and method definition is  behind DEBUG flag. 

If we go back to the prev implementation of calling c function, then that would be better.

react-native/react-native-flipper/react_native_flipper.rb
```diff
@@ -0,0 +1,44 @@
...
+    if target.name == 'YogaKit'
+      target.build_configurations.each do |config|
+        config.build_settings['SWIFT_VERSION'] = '4.1'
+      end
+    end
+  end
+end
+
+
+def use_flipper!(versions = {}, configurations: ['Debug'])
```
priteshrnandgaonkar, line 12: Why do we have to define this ? Shouldn't  this come from RN core ? Are you suggesting  to move this configuration in Flipper ? cc  @mweststrate

priteshrnandgaonkar (changes requested):
I have few suggestions and a questions

react-native/react-native-flipper/react_native_flipper.rb
b3coded: Yes, currently it's coming from RN core.
my suggestion is to move this to Flipper to be able to change this without needing to upgrade inside the RN core

react-native/ReactNativeFlipperExample/ios/ReactNativeFlipperExample/AppDelegate.m
b3coded: I'm not familar with Obj-C ecosystem so I'm prepared to change this if needed, this is just a initial implementation of this change I'm proposing.

But you can keep the current static InitializeFlipper inside your `AppDelegate.m` as a custom Flipper initialization

react-native/ReactNativeFlipperExample/ios/ReactNativeFlipperExample.xcodeproj/project.pbxproj
b3coded: It's coming from 'react_native_post_install' script
the RN 0.64.x uses this function on Podfile but I think someone missed this

react-native/react-native-flipper/react_native_flipper.rb
priteshrnandgaonkar: Then I have few questions. How will we ensure that bumping up the deps in this function won't break RN master and the scheduled release of RN, thats why I feel its better that this configuration is part of RN core rather than here. 

Library/Framework releases shouldn't be worried about the clients integrating and if a client builds(ofcourse we have our own flipper CI in place to avoid breakages) or not, its rather clients  responsibility to verify if they are compatible with latest library/framework release version and its them who should take the call to bump rather than us. Isn't it ?

@mweststrate  what do you think ?

react-native/ReactNativeFlipperExample/ios/ReactNativeFlipperExample/AppDelegate.m
priteshrnandgaonkar: So can you revert this change ?

react-native/react-native-flipper/react_native_flipper.rb
b3coded: Yes, you are correct, but in this PR I'm implementing on `react-native-flipper` suggesting to this library to be an integration with Flipper. The idea is this dep should take care of how to make it work with RN.

Since `react-native-flipper` is an integration with custom Flipper Plugins, I think it should be able to handle not only custom plugins but the Flipper integration itself. It ma be an more easy way to manage Flipper issues on RN side.

The plan if the PR is merged is to change the default template to include the `react-native-flipper` dep by default using the default initialization methods that are already in this dependency, but it can be customized on project side (like the current RN template already does)

mweststrate: @b3coded I think it is a good idea to first raise this as a proposal discussion (RFC) on the react-native repo itself, so that folks over there are involved as well. I'm still a bit unclear what problems this solves:

1. customizing the flipper version, or removing it, is all done in the podfile of your repository with the `use_flipper` command. I'm not sure how that is made easier over here?
2. by no longer being part of the RN core setup, signal on issues will arrive much later


So, before diving into the technical correctness of the solution / code, I'd love to better understand the proposal better first? You state  "This takes advantage of the react-native-flipper dependency to make possible more easy way to install, remove and update Flipper without needing to change the react-native repo.", but I'm missing the how / why of that statement :). Could you show how a typical react native project looks before and after the changes proposed? And how the impact of bumping a Flipper version is different? And how the compatibility between Flipper and RN would still be guaranteed by CI? This might be a very good proposal, but so far I don't see enough rationale yet :).

b3coded:
Thanks for the reviews @mweststrate @passy @priteshrnandgaonkar
I will try to add more details to this proposal in an RFC

mweststrate:
Closing this PR for inactivity, can be reopened if needed

PR #2574, opened 2021-07-07
Fix selection on mock list

bizzguy:
## Summary

Fix selection problem on ManageMockResponsePanel.  Currently, when the user adds new routes, the detail section continues to show the detail for the first route.  Select different routes in the list does not change the detail panel.  The screen is currently not doing anything on selection.

Also fixed a problem with updating on mock header names and values.  Updates were not persisting if user did not exist input fields before leaving header input panel.

## Changelog

Network Plugin - Fix selection problem on ManageMockResponsePanel

## Test Plan

Add multiple mock routes
Select various routes
Verify that detail panel shows the selected route


mweststrate:
Thanks for picking this up @bizzguy!

PR #2704, opened 2021-08-18
Revert "Move UnitializedClient to server"

passy:
This reverts commit 0cbe063c77500f607ad653cced8a3782568bc816.

Attempt to fix https://github.com/facebook/flipper/runs/3351826251?check_suite_focus=true


mweststrate:
More likely that it was the parent commit that broke this, I'll investigate

passy:
> More likely that it was the parent commit that broke this, I'll investigate

Yep. Shows in the signals here, too.

Issue #2708, opened 2021-08-19
Discussion: Automatic Dark Theme


swrobel:
Now that the dark theme is as usable as the light one (thanks!) it would be nice to have the theme automatically switch based on the system theme. In the previous dark theme discussion, someone [pointed out the Electron API](https://github.com/facebook/flipper/issues/552#issuecomment-533111444) for this.

jknoxville:
Makes sense to me, I guess we want to change the settings from:

`Enable dark theme: Yes | No`

to some 3 way switch:

`Enable dark theme: Yes | No | Use system setting`

With "Use system setting" being the new default.

qichuan:
@passy https://github.com/facebook/flipper/pull/2759 Here is the PR :)

mweststrate:
Implemented, so closing. Thanks for the PR @qichuan!

PR #2759, opened 2021-08-31
Add option for automatic dark theme

qichuan:
## Summary

Add an setting option to allow automatic dark theme

A feature requested by @swrobel https://github.com/facebook/flipper/issues/2708

### UI change
Replace the `Enable Dark Theme` Toggle button with a Radio group containing three Radio buttons. 

The available options are `Dark`, `Light` and `Use System Default`, of which `Use System Default` is the default value

Screenshot 2021-08-31 at 3 32 44 PM

### Data structure change

The `darkMode` property of [Settings](https://github.com/facebook/flipper/blob/c0cd32564ad4b07a42365af886045095877a6a6b/desktop/app/src/reducers/settings.tsx#L20) is changed from `boolean` to `string`, the available values are `dark`, `light` and `auto`

### Test 1
- Step: Choose `Dark` in the Settings page and click on `Apply and Restart` button
- Expect: The application is displayed in Dark mode
### Test 2
- Step: Choose `Light` in the Settings page and click on `Apply and Restart` button
- Expect: The application is displayed in Light mode
### Test 3
- Step: Choose `Use System Default` in the Settings page and click on `Apply and Restart` button
- Expect: The application is displayed in System Default mode

### DIFF
--- a/desktop/app/src/init.tsx
+++ b/desktop/app/src/init.tsx
@@ -225,7 +225,7 @@ function init() {
       }
       return {
         shouldUseDarkMode: shouldUseDarkMode,
-        theme: theme
+        theme: theme,
       };
     },
     (result) => {
--- a/desktop/app/src/init.tsx
+++ b/desktop/app/src/init.tsx
@@ -49,7 +49,7 @@ import styled from '@emotion/styled';
 import {CopyOutlined} from '@ant-design/icons';
 import {getVersionString} from './utils/versionString';
 import {PersistGate} from 'redux-persist/integration/react';
-import {ipcRenderer} from 'electron';
+import {ipcRenderer, remote} from 'electron';
 
 if (process.env.NODE_ENV === 'development' && os.platform() === 'darwin') {
   // By default Node.JS has its internal certificate storage and doesn't use
@@ -213,9 +213,20 @@ function init() {
   sideEffect(
     store,
     {name: 'loadTheme', fireImmediately: false, throttleMs: 500},
-    (state) => ({
-      dark: state.settingsState.darkMode,
-    }),
+    (state) => {
+      const darkMode = state.settingsState.darkMode;
+      let shouldUseDarkMode = remote.systemPreferences.isDarkMode();
+      if (darkMode === 'dark') {
+        shouldUseDarkMode = true;
+      } else if (darkMode === 'light') {
+        shouldUseDarkMode =  false;
+      } else if (darkMode === 'auto') {
+        shouldUseDarkMode = remote.systemPreferences.isDarkMode();
+      }
+      return {
+        dark: shouldUseDarkMode
+      }
+    },
     (theme) => {
       (
         document.getElementById('flipper-theme-import') as HTMLLinkElement
--- a/desktop/app/src/init.tsx
+++ b/desktop/app/src/init.tsx
@@ -219,13 +219,13 @@ function init() {
       if (darkMode === 'dark') {
         shouldUseDarkMode = true;
       } else if (darkMode === 'light') {
-        shouldUseDarkMode =  false;
+        shouldUseDarkMode = false;
       } else if (darkMode === 'auto') {
         shouldUseDarkMode = remote.systemPreferences.isDarkMode();
       }
       return {
-        dark: shouldUseDarkMode
-      }
+        dark: shouldUseDarkMode,
+      };
     },
     (theme) => {
       (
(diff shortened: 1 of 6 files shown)


passy (approved):
Amazing stuff! Thanks for building this!

desktop/app/src/utils/useIsDarkMode.tsx
```diff
@@ -8,12 +8,23 @@
...
 /**
  * This hook returns whether dark mode is currently being used.
  * Generally should be avoided in favor of using the above theme object,
  * which will provide colors that reflect the theme
  */
 export function useIsDarkMode(): boolean {
-  return useStore((state) => state.settingsState.darkMode);
+  return useStore((state) => {
+    const darkMode = state.settingsState.darkMode;
+    if (darkMode === 'dark') {
```
passy, line 21: maybe `|| darkMode === true` for backward compatibility

desktop/app/src/utils/useIsDarkMode.tsx
```diff
@@ -8,12 +8,23 @@
...
+      return true;
+    } else if (darkMode === 'light') {
+      return false;
+    } else if (darkMode === 'auto') {
+      return remote.systemPreferences.isDarkMode();
+    }
+    return remote.systemPreferences.isDarkMode();
```
passy, line 28: Let's keep `light` as the default

```
return false
```

desktop/app/src/init.tsx
```diff
@@ -213,9 +213,20 @@ function init() {
   sideEffect(
     store,
     {name: 'loadTheme', fireImmediately: false, throttleMs: 500},
-    (state) => ({
-      dark: state.settingsState.darkMode,
-    }),
+    (state) => {
```
passy, line 216: Mac switches dynamically between light and dark based on time (if set up), so I guess we have to set some system listener for that to reflect that, in case the mode is auto?

passy (changes requested):
I'm relaying some internal feedback I got on the diff. :)

desktop/app/src/init.tsx
swrobel: The code 2 lines down is using the Electron API to tap into this, but is it not listening correctly to changes to update dynamically when the system switches?

`let shouldUseDarkMode = remote.systemPreferences.isDarkMode();`

qichuan: Maybe we can subscribe the `AppleInterfaceThemeChangedNotification` event and refresh the UI when the system theme changes? 

But my concern is that the UI refresh could be interruptive, is there a way to update the application theme without a refresh?

desktop/app/src/utils/useIsDarkMode.tsx
qichuan: OK, no problem

desktop/app/src/init.tsx
mweststrate: I think the process being interruptive is fine, I see more applications doing that. I think the clue is to listen to the events exposed here https://www.electronjs.org/docs/api/native-theme in the flipper entry file, roughly following https://www.electronjs.org/docs/tutorial/dark-mode (which is partially already implemented). Updating the setting in the Redux store when the updated event happens, should apply the change automatically without needing a restart.

qichuan: I have replaced the `systemPreferences.isDarkMode()` with `nativeTheme.shouldUseDarkColors`. The application can now automatically react to the system theme changes without the need of UI refresh :)

@passy could you please review my changes again? Thanks.

mweststrate:
Looks like the unit tests are failling, would you mind running `yarn watch` and update the snapshots?

qichuan:
> Looks like the unit tests are failling, would you mind running `yarn watch` and update the snapshots?

I am getting this from `yarn watch`

```
Summary of all failing tests
 FAIL  app/src/server/devices/ios/__tests__/iOSDevice.node.tsx
  ● Test suite failed to run

    Call retries were exceeded

      at ChildProcessWorker.initialize (node_modules/jest-worker/build/workers/ChildProcessWorker.js:193:21)


Test Suites: 1 failed, 1 skipped, 93 passed, 94 of 95 total
Tests:       2 skipped, 531 passed, 533 total
Snapshots:   203 passed, 203 total
Time:        73.712 s
Ran all test suites.
```

Not sure if it is caused by my PR... @mweststrate

mweststrate:
Doesn't sound like it... will check next week!

Issue #2896, opened 2021-09-21
Xcode 13 - developer tools 


sramezani:
## 🐛 Bug Report
I tried to work with Flipper V 0.110.0 and Xcode V 13, but when I opened the Flipper I faced with this box:

<img width="456" alt="Screenshot 2021-09-21 at 15 45 05" src="https://user-images.githubusercontent.com/33284430/134172747-ca0d7923-adce-4084-a24f-5387901ed701.png">

I tried to install multiple times, removed Flipper and installed again, downloaded developer tools directly from here https://developer.apple.com/download/all/?q=command%20line
but doesn't work, and show me this message again and again.

## Environment
Flipper: v0.110
Xcode: 13.0
macOS: 11.6
react-native: 0.65.1

carloscrovara:
Same issue. Any solutions?

samuel29k:
I've done some investigation and  this seems to be caused by the new command line tools not registering correctly in the path / or not getting installed properly.

Could this be related to the path where the Xcode command line tools are installed?

It seems like running `xcode-select -p`  in a terminal yields `/Applications/Xcode.app/Contents/Developer` after installing Xcode 13.0.

However, before the Xcode 13 upgrade, `xcode-select -p` yielded `/Library/Developer/CommandLineTools` for me.

webdevbyjoss:
So it looks like it should be as simple as replacing the `instruments -s devices` call with `xcrun xctrace list devices`, right?
And maybe tweak the output processing a bit to skip the unnecessary headers.

webdevbyjoss:
Yup, it works with that change! 
![Screenshot 2021-09-23 at 19 35 51](https://user-images.githubusercontent.com/402001/134548689-26bb477c-3668-48f1-a7fb-396287daabed.png)

VytautasLozickas:
A quick and dirty fix while we wait for the new Flipper release - add this to your `~/.bashrc` or `~/.zshrc`:
`alias -g "instruments"="xcrun xctrace list devices #"`

sramezani:
Issue fixed on version 0.113.0

PR #2903, opened 2021-09-23
Migrate away from deprecated instruments utility

webdevbyjoss:
## Summary

The "instruments" utility was deprecated in Xcode 12 and removed now from Xcode 13, causing all the iOS 15 / Xcode 13  builds to fail to work with Flipper.

See deprecation notice here: https://developer.apple.com/documentation/xcode-release-notes/xcode-12-release-notes

> The instruments command is now deprecated in favor of its replacement: xctrace. Use xctrace to record, import, and export data from Instruments .trace files. (36641078)

The corresponding bug was reported here:
https://github.com/facebook/flipper/issues/2896

That call should be replaced with `xcrun xctrace` run and that is what we are doing here.

IMPORTANT: There was a workaround on killing the stale `instruments` proceses and that utility function is removed as part of this PR, as `instruments` were slow and inefficient, marked for deprecation more than a year ago. `xctrace` should not have that kind of issue and it doesn't make sense to keep this workaround now. 

## Changelog

The "instruments" call replaced with `xcrun xctrace` use, to align with deprecation introduced in Xcode 12 and offer Xcode 13 compatibility.

## Test Plan

In order to test this we should run the Flipper desktop app and make sure it connects to the Simulator started out of Xcode 13
It fails to do so in `master` version right now, see https://github.com/facebook/flipper/issues/2896


passy:
This looks great! Thanks a lot for changing that. I wasn't aware of the replaced tool. We will probably need to do a bit of testing internally before shipping this.

webdevbyjoss:
@passy can you plz give me a hint on what could be the issue with the failing tests? I'm new to the code and don't see anything in my changes that could break the build here

Tried running those locally and they fail on `master` as well, so it looks like something not related to the change we have in this PR.

sebazelonka:
Any news on this?

webdevbyjoss:
I'm not sure what is the process after the pull request was imported. Someone who is a Facebook employee could take a look at Phabricator and check the progress there. 

I'd say we should consider this issue as **critical**, as it blocks Flipper from being used with Xcode 13 which is the latest and greatest version right now. People from all around the world can't migrate their React Native apps to iOS 15 due to this, unless they drop Flipper for now.

mweststrate:
We're currently testing internally, should ideally end up in tomorrows release.

passy:
There are a few more changes we need to make. For instance, the Doctor needs to be updated to skip the `instruments` check. I hope I'll be able to do that today and then we can merge this along with the other changes. :)

m-jachym:
image

PR #2904, opened 2021-09-23
[actions] Create issue on publish failure

passy:
## Summary

Because the publish runs aren't triggered on the main branch, we have limited visibility. Sending push notifications, webhooks or email integrations are hard, but opening an issue seems like a decent work-around for this.

We can broaden this to more publish actions as needed.

## Test Plan

Ran on my fork which generated this issue: https://github.com/passy/flipper/issues/9


passy:
Rebasing so the `main` errors go away.

PR #3041, opened 2021-11-11
Updating splitting of avd names to work on Windows and Unix based sys…

rasaha91:
…tems

## Summary

When trying to start an android emulator in Windows, there is a failure in trying to read `<avd name>.ini`. The reason for this is because when we run `<android sdk path>\tools\emulator.exe -list-avds`, a split is performed with the new line character (`\n`). Unfortunately, this does not work for Windows, which uses `\r\n` as the new line termination. This results in a carriage return remaining on each output string, which results in trying to access `<avd name> .ini`, which is an invalid path.

The fix is to update the split to use a regex that is compatible for both Unix and Windows based systems.

## Changelog

Updating splitting of avd names to work on Windows and Unix based systems

## Test Plan

Verified the avd names didn't have any trailing whitespace, and the emulator is able to launch.

### DIFF
--- a/desktop/flipper-server-core/src/devices/android/androidDeviceManager.tsx
+++ b/desktop/flipper-server-core/src/devices/android/androidDeviceManager.tsx
@@ -149,7 +149,7 @@ export class AndroidDeviceManager {
             return;
           }
           const devices = data
-            .split('\n')
+            .split(/\r?\n/)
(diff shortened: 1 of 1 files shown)


nikoant:
Thanks for PR, @rasaha91, I'll take care to check it with internal CI

Issue #3117, opened 2021-11-29
🐛 MacCatalyst Does not Build on RN above 64 because of RCT-Folly (1/2 Fix Included Here!)


ZComwiz:
## 🐛 Bug Report

**FLIPPER FOLLY MAC CATALYST SUPPORT BROKE AFTER SWITCHING TO XCFRAMEWORK. HERE IS WHAT HAPPENED:**
There are two bugs when building RN 65+ targeted to MacCatalyst located in Flipper:

**1. Time.h redefinition.** There is a hack in [here in Time.h](https://github.com/facebook/folly/blob/main/folly/portability/Time.h#L31) in RCT which I was able to consistently fix by adding a TARGET_OS_MACCATALYST flag on line 31. See my comment on the quick, robust, fix [here](https://github.com/reactwg/react-native-releases/discussions/1#discussioncomment-1711630).

**2. DoubleConversion Not Found** when building:
When Flipper/Folly switched to xcFrameworks from the precompiled .a file, it seems they did not include a slice for MacCatalyst. I will include some helpful references here which the team at Flipper who is more familiar with the archiving process of the xcframework can use to quickly rectify this support regression:

a) [Mike Hardy's awesome script which may help compile the DoubleConversion Library properly](https://github.com/transistorsoft/react-native-background-fetch/issues/311#issuecomment-774781786)

b) A [simple guide to building MacCatalyst slices](https://pspdfkit.com/blog/2020/supporting-xcframeworks/) when using an xcFrameworks

## To Reproduce

Try building any react-native on any version 65 or greater with Mac as the target destination (MacCatalyst). You can do that by selecting iPad, then Mac in the General settings of the template HelloWorld Xcode Project. Switch Build Target to Mac at the top.

Recommend using the template in[ this branch of a release candidate of 67](https://github.com/facebook/react-native/pull/32633) which adds some robust fixes that will be integrated soon (but you can try on the latest RN 66 build).

## Environment

Intel and M1 Macs alike: targeting MacCatalsyt. XCode Version 13.0 (13A233). Any version of React Native above 64, but you can use 66 or the release candidate mention above.

mikehardy:
Wow, that took a while to find.
It's here and needs attention from @priteshrnandgaonkar 

https://github.com/priteshrnandgaonkar/double-conversion/blob/master/archive-xcframework.sh

Pritesh, you need to include a macCatalyst slice, easy to build but needs a very very careful re-arrangement of built artifacts after build, before packaging:

[...]

PR #3120, opened 2021-12-01
Update to enable Mac universal builds of Flipper

colinbendell:
## Summary

New M1 Mac's must use rosetta to use Flipper. The Electron builder only produces x86_64 builds and the PortForwardingApp is precompiled and committed in the code base with an x64 build. This PR adds universal building of the electron app and the PortForwardingMacApp

## Changelog

Enable universal mac (mac/x64) builds

## Test Plan

* `yarn build --mac` to build the application as per usual
* run the flipper.app as usual
* verify that app is working and that on an m1, the `Kind` is `Apple`


iOS/Podfile
```diff
@@ -19,3 +20,11 @@ target 'FlipperKit' do
   pod 'SocketRocket', '~> 0.6.0'
 
 end
+target 'PortForwardingMacApp' do
+  #use_frameworks!
+  platform :osx, '11'
+  workspace 'FlipperKit.xcworkspace'
+  project '../PortForwardingMacApp.xcodeproj'
```
colinbendell, line 27: Placing the PortFrowarding project at the root level is messy. I only placed it there because the src is split between `/PortForwardingMacApp/main.m` and `/ios/FlipperKit/FKPortForwarding/*`. 

Q: should we just refactor this and place the project and relevant sources under `/desktop`?

colinbendell: Decided to refactor anyway. PR now moves `/ios/FlipperKit/FKPortForwarding/FKPortForwardingClient.*` to `/desktop/PortForwardingMacApp`

passy:
Thanks for putting this together, @colinbendell!

We'll probably need some time to look into this. I'm not quite sure what this means when rolling it out to M1s due the binary signing requires for which we currently have no infrastructure in place.

colinbendell:
> We'll probably need some time to look into this. I'm not quite sure what this means when rolling it out to M1s due the binary signing requires for which we currently have no infrastructure in place.

An (interim) solution could be to lean on homebrew: change the formula to build flipper on the local device, and use self-signed certs. This would solve the arm64e requirement and eliminate the rosetta dependency while you sort out the code signing requirements for the infrastructure more broadly.

colinbendell:
Why not simply use ad-hoc signed binaries? This is the approach that [homebrew uses](https://github.com/Homebrew/brew/pull/9102) instead of signing binaries.

I've confirmed that with xcode 13 and the latest `electron-builder` that the resulting binaries are already ad-hoc signed when a valid Identity is not found or when `CSC_IDENTITY_AUTO_DISCOVERY=false`. Using this as-is would be sufficient for the time being.

marko-asplund:
Any news on publishing universal builds?

ZComwiz:
Let's not leave Mac Catalyst behind: https://github.com/facebook/flipper/issues/3117

colinbendell:
> Let's not leave Mac Catalyst behind: #3117

what's the intersection with this PR and Mac Catalyst?

aarongrider:
Let me know if I can help at all with getting this over the line. Looks like we might need a rebase.

umangloria:
Still waiting for this PR to get merged  :(

lockieluke:
please hurry up react native devs apple silicon machines were released two years ago

cristiandan:
i guess this is just as good as https://github.com/facebook/flipper/pull/3553

Issue #3173, opened 2021-12-16
Discussion: Syntax Highligting for Logs Flipper Plugin


clytras:
Hello,

For productivity, I wanted syntax highligting for the console logs and this is/was something that kept me from using Flipper and relying to a browser debugger console for debugging.

I gave it a shot and I managed to make Flipper Logs plugin to have syntax highlighting (_with `TruncateHelper` even_) using [`react-syntax-highlighter`](https://www.npmjs.com/package/react-syntax-highlighter) which supports both [highlight.js](https://highlightjs.org/) and [Prism.js](https://prismjs.com/).

You can see the result using dark mode and _Night Owl_ [highlight.js theme](https://highlightjs.org/static/demo/):

![image](https://user-images.githubusercontent.com/6103077/146371127-e3ece4f7-f458-4699-8c05-b243073bf686.png)

It's not perfect as this method highlights everything, every message, but the majority of the massages (at least for me) are code references and simple strings are printed with single color except if they have some javascript keywords that get highlighted; even though, it's way better like this than not having syntax highlighting at all.

If you guys like it, I can provide a PR for Logs plugin as a starting point to make Flipper support syntax highlighting for _Logs_.

efstathiosntonas:
For the name of God, please submit this PR! Thanks Chris!

mweststrate:
Looks cool! My primary concern would be around performance, logs plugin can generated a few hundred K of lines and scroll very fast, so I'd check against a physical android device's logs to see if this doesn't slow down rendering a lot. If it does, maybe it could be made an option in the drop down menu?

clytras:
@efstathiosntonas Hey Stathis! How nice to see you here! We need to get in touch!

I wanted to discuss this first before I procced submitting a PR.

@mweststrate I share the same concerns with you regarding performance, but remember, for each record, the highlighted text will be the truncated text, which currently has a length of 400:

https://github.com/facebook/flipper/blob/d48fbd8e50c8ae6203a8ce92aa5a98aa64160ca1/desktop/plugins/public/logs/index.tsx#L100-L104

so, even id there is a 5K string/object, the highlighter will always process the first 400 characters, it will only process the whole log entry when the user hits the *more* expand button.

It works pretty well so far on a big RN project with simple logs.

This definitely needs some options to the settings, not only for toggling its' state, but setting up the theme as well and I think it would be best to flag it (*and actually display it*) as experimental and even warn users that they may experience performance issues on big log loads/usage.

[...]

PR #3226, opened 2021-12-31
Add syntax highlighting for logs

clytras:
This a follow up PR of the https://github.com/facebook/flipper/issues/3173 discussion and is a staring point for adding syntax highlighting for logs plugin.

## Summary

Syntax Highlighting helps with productivity, and it's something that the logs on Flipper currently lack. As I mention on the discussion OP, this is not perfect as it will highlight everything, but (*at least for me*) it's better to highlight everything than nothing! And, in most cases, simple string messages will be displayed highlighted with a single color.

## Approach

I'm implementing all syntax highlighting inside `desktop\flipper-plugin\src\ui\DataFormatter.tsx` by creating a `highlightSyntax` formatter. The caviat with this approach, is that `truncate` formatter can't be processed, because `truncate` formatter returns JSX if value exceeds `maxLength` and `highlightSyntax` formatter can only process a string, not JSX; thus, I have to implement SH and inside `TruncateHelper` component in order to make it highlight truncated values. The `language` prop of the `SyntaxHighlighter` component is hardcoded to `javascript` and I think this is the most suitable value that matches DevTools console debugger.

## Settings

- There is a new toggled setting named `enabledLogsSyntaxHighlighting` and labeled "Syntax Highlighting for Logs (experimental)" that will enable/disable the feature. It's turned off by default.
- There is a new select setting named `logsSyntaxHighlightingStyle` and labeled "Highlighting style" to select Highlight.js theme/styles (https://highlightjs.org/static/demo/) to apply to the highlighted values. This loads all available styles from inside `react-syntax-highlighter/dist/cjs/styles/hljs` (used CJS instead of ESM because I'm having issues with Jest) and it presents all available styles by normalizing camel case key names using a new utility function `normalizeCamelCase` from `flipper-ui-core/utils/normalizeCamelCase`. The default value is `atomOneLight` since the app uses default light mode. It will also use the default style (`atomOneLight`) if the setting value is falsy.

## PR issues

1. Typings won't pick for `react-syntax-highlighter/dist/cjs/styles/hljs`; had to `ts-ignore` it. I've literaly spent hours to make this work with `@types/react-syntax-highlighter`, but I don't have much experience working with yarn workspaces and typescript typings.
1. How to handle default value for `logsSyntaxHighlightingStyle`. Now, I've hardcoded the default style (`atomOneLight`), but that's not ideal. Maybe I should export a `flipper-common` constant with the default value?
1. Linting: I'm having 73 warnings when running `yarn lint` (`'flipper' import is restricted from being used. Direct imports from 'flipper' are deprecated`) and an error (`Error: Command failed: ./node_modules/.bin/tsc -p ./plugins/tsconfig.json` `'.' is not recognized as an internal or external command, operable program or batch file`) but I think all of these are not related to my edits.
1. I cannot figure out the best way to access the settings inside `flipper-plugin/src/ui/DataFormatter.tsx`. I created a `global.ts` declaring `global { Window { FlipperRenderHostInstance } }` and then accessing settings using `window.FlipperRenderHostInstance.serverConfig.settings`.

## Side fix

I also noticed that there is a bug on settings dialog when hitting "Apply and Restart" button, the settings would not have picked after the restart, and I think that's because the process gets restarted before `flush` promise resolves. I changed `applyChanges` to first `await flush()` and then restart Flipper and now it works.

## Changelog

New experimental feature for supporting logs syntax highlighting.

## Test Plan

I haven't add any tests and all tests do pass because the setting is turned off by default, so Jest snapshots will always match existing ones.

### Screenshots

Screenshot on Light mode with "Atom One Light" style

image

Screenshot on Dark mode with "Night Own" style

image

Settings screenshot

image

Tests result

image


aigoncharov:
@clytras thank you for your PR! Could. you rebase it to the latest `main`?

clytras:
@aigoncharov thanks for the follow up.
I just did rebase but I'm not sure if I did it correctly!
Please let me know if there are any issues.

desktop/flipper-plugin/src/ui/DataFormatter.tsx
```diff
@@ -187,3 +198,31 @@ const truncateButtonStyle = {
...
+
+type SyntaxHighlightProps = {
+  children: React.ReactNode;
+};
+
+function SyntaxHighlight({children}: SyntaxHighlightProps) {
+  const {
+    enabledLogsSyntaxHighlighting,
+    logsSyntaxHighlightingStyle = 'atomOneLight',
+  } = window.FlipperRenderHostInstance.serverConfig.settings;
```
aigoncharov, line 212: Instead of using RenderHost directly, we expose [`FlipperLib`](https://github.com/facebook/flipper/blob/main/desktop/flipper-plugin/src/plugin/FlipperLib.tsx) to plugins. 
Could we extend it instead?
```ts
interface FLipperLib {
  logsSyntaxHighlighting?: {
    enabled: boolean,
    style: keyof HljsStylesType
  }
}
```
We could pass settings from render host to flipper lib [here](https://github.com/facebook/flipper/blob/main/desktop/flipper-ui-core/src/utils/flipperLibImplementation/index.tsx#L39)

aigoncharov: I am also wondering if we should check it in `highlightSyntax` and remove the checks from the component itself?

aigoncharov:
> @aigoncharov thanks for the follow up. I just did rebase but I'm not sure if I did it correctly! Please let me know if there are any issues.

Great job! Left a couple of comments, but it looks really promising! Thank you for working on it!

desktop/flipper-plugin/src/ui/DataFormatter.tsx
clytras: Thank you for your review @aigoncharov. I have the checks inside the component, because there we need to get the setting for the selected highlighting theme, and I thought, since we're going to use the settings there, why not check for both *theme* and *enabled* flag. I'll put it inside the formatter instead and leave the theme setting inside the component; or maybe fetch *theme* setting even inside the `highlightSyntax` formatter and pass it as a prop to the `SyntaxHighlight` component? Maybe that's even better.

clytras: I'll also try to extend `FlipperLib` as you suggested and update here. Hope I'll manage to address all of your points today to forward this PR.

PR #3328, opened 2022-01-23
[desktop] fix ReleaseChannel info in WelcomeScreen in dark mode

Simek:
## Summary

This PR fixes the color of ReleaseChannel information in the WelcomeScreen, when user uses dark mode.

I have decided to switch to `textColorSecondary` instead of `textColorPrimary` to emphasise a bit more the information, but I can change it to the primary color if you think it will fit there better. Also the font size has been adjusted to match the paragraph font size.

## Changelog

* fix ReleaseChannel info in WelcomeScreen in dark mode

## Test Plan

#### Dark Mode (before & after)

Screenshot 2022-01-23 at 15 14 12
Screenshot 2022-01-23 at 15 09 58

#### Light Mode
Screenshot 2022-01-23 at 15 22 42

### DIFF
--- a/desktop/flipper-ui-core/src/sandy-chrome/WelcomeScreen.tsx
+++ b/desktop/flipper-ui-core/src/sandy-chrome/WelcomeScreen.tsx
@@ -141,9 +141,7 @@ export function WelcomeScreenStaticView() {
 }
 
 function WelcomeScreenContent() {
-  function isInsidersChannel() {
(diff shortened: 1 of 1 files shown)


Simek:
@lblasa I have just spotted the ESLint issues with the changed lines, should I update the PR to fix those?

I have skipped this checks earlier since ESLint was reporting +60 issues on local `main` checkout and I don't want to introduce the unrelated changes. I can also prepare separate PR which fixes those issues, let me know what is the preferred approach.

PR #3329, opened 2022-01-23
[desktop] do not add `GrayedOutOverlay` if there is no children

Simek:
## Summary

Do not add `GrayedOutOverlay` to the `ToggledSection` component if there is no children.

Current approach was causing UI issue in the Settings Modal in dark mode.

## Changelog

* do not add `GrayedOutOverlay` to the `ToggledSection` if there is no children

## Test Plan

## Preview (before & after)

Screenshot 2022-01-23 at 15 32 29
Screenshot 2022-01-23 at 15 32 12


passy:
Thanks for the fix! :)

PR #3372, opened 2022-01-30
[desktop] UI: fix hover effect of LeftRail icons with badge

Simek:
## Summary

This PR fixes the missing hover effect for the `LeftRail` component buttons, when they have included the badge.

To fix the issue, I had to wrap the whole `Button` with `Badge` component (instead wrapping only around icon). However, this solution required to added `offset` property to the `Badge` which moves the indicator to the position prior change (otherwise indicator was moved to the right corner of the button).

The file has been formatted after the changes with ESLint.

**Edit:** I have also spotted that this change fixes the icon placement inside the button, when badge is present. Earlier, as seen below, the log icon was moved towards the top of the button:
Screenshot 2022-01-31 at 00 57 49

## Changelog

[desktop] UI: fix hover effect of LeftRail icons with badge

## Test Plan

#### Before
Screenshot 2022-01-31 at 00 24 23

#### After

> I have used mocked indicator values locally to ensure that the Badges are displayed correctly.

Screenshot 2022-01-31 at 00 26 10

Screenshot 2022-01-31 at 00 45 08


Simek:
I have missed that after the change clicking Badge do not fire the button. I have two ideas how to fix that, and I will investigate those soon.

Simek:
I have tried to fix this by altering the DOM hierarchy and also by moving icon for those buttons to the `children`, but every solution I have tried was causing another problem. 

After a bit of experimenting, it looks like the most straight forward solution is just to add `onClick` also to the `Badge` component, however this triggers TS waring due to mismatched types (this is why I have opted to pass the prop in that way, to avoid `@ts-ignore`), in other words `Badge` component do not specify that this prop can be passed:
* https://ant.design/components/badge/#API

But it works as expected, so I'm not sure if it's fine to merge as is, or this problem need a deeper dive into `antd` code and PR on their side.

PR #3392, opened 2022-02-02
Add support for "Application Support" folders. Fixes #3391

### LINKED ISSUE #3391: Flipper cannot connect to React Native apps run on My Mac (Designed for iPad)
## 🐛 Bug Report
If you want to test the react native app to see how it runs on an M1 Mac, flipper currently can't connect to it.

The error thrown is

`Path didn't match expected pattern: /Users/andreibarabas/Library/Containers/A8D9392D-0686-4C6A-AFC0-7AAF83B1EDF1/Data/Library/Application Support/sonar/app.csr`

## To Reproduce
In the Xcode simulator list, select My Mac (Designed for iPad). It needs to be an M1 Mac

## Environment
Flipper Desktop: v0.131.1
MacOS: 12.1
Xcode: 13.2.1

andreibarabas:
This is the case when running a mobile app directly on an M1 machine. Fixes #3391

## Summary

You can't currently connect to an app that runs directly on the M1 (not a simulator)

## Changelog

Support for apps running directly on an M1 machine (not via a simulator)

## Test Plan

In Xcode, in the simulator dropdown, select the M1 Mac and run the app

### DIFF
--- a/desktop/flipper-server-core/src/utils/CertificateProvider.tsx
+++ b/desktop/flipper-server-core/src/utils/CertificateProvider.tsx
@@ -255,9 +255,9 @@ export default class CertificateProvider {
   }
 
   private getRelativePathInAppContainer(absolutePath: string) {
-    const matches = /Application\/[^/]+\/(.*)/.exec(absolutePath);
-    if (matches && matches.length === 2) {
-      return matches[1];
+    const matches = /Application( Support)?\/[^/]+\/(.*)/.exec(absolutePath);
+    if (matches && matches.length === 3) {
+      return matches[2];
     }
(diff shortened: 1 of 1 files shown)


mweststrate:
cc @lawrencelomax would you mind reviewing this one? I imagine we probably need to support both cases?

lawrencelomax:
@andreibarabas I don't have an M1 on me and I don't quite understand where things "moved to" on M1s.

Can you please describe the x86 vs M1 directory changes? I suspect we need to unit test this function to be sure that it handles the x86 vs M1 case. My expectation here is that on M1s a simulator has it's files within `Application Support` and `Application` x86 hosts.

The regex does look fine to my eyes as ` Support` is effectively optional.

PR #3398, opened 2022-02-05
Change label from iOS to ObjC

tbskgm0:
## Summary
The documentation shows Swift and ObjC code, but the label names are Swift and iOS.

## Changelog
Modify the documentation

## Test Plan
No test plan


docs/getting-started/ios-native.mdx
```diff
@@ -89,7 +89,7 @@ This is done to overcome a bug with Xcode 11 which fails to compile swift code w
 Install the dependencies by running `pod install`. You can now import and initialize Flipper in your
 AppDelegate.
 
-<Tabs defaultValue="ios" values={[{ label: 'iOS', value: 'ios'}, { label: 'Swift', value: 'swift'}]}>
+<Tabs defaultValue="ios" values={[{ label: 'ObjC', value: 'ObjC'}, { label: 'Swift', value: 'swift'}]}>
```
aigoncharov, line 92: Should we also change `defaultValue`?

aigoncharov: CC @tbskgm0

tbskgm0: Yes, that's right.
It's strange that the defaultValue is iOS when the values are objC and swift.
In the following docs, the defaultValue is objC. So, I think the defaultValue should be set to objC again.
[<Tabs defaultValue="objc" values={[{ label: 'ObjC', value: 'objc'}, { label: 'Swift', value: 'swift'}]}>](https://github.com/facebook/flipper/blob/main/desktop/plugins/public/network/docs/setup.mdx)

Issue #3320, opened 2022-01-19
Discussion: 3rd party NPM registry, internal only plugins


usrbowe:
I would like to open discussion on this topic. Most plugins will likely be open sourced and published to `npm`. But I believe lot of use cases will include internal plugins, which should only be available internally.

I guess this is also case in Facebook/Meta. So would like to know what approach you guys took, and maybe we could add some similar option for others as well?

One of the main issues, might be how to actually add 3rd party registry, since the searching through registry might be different and also parsing results will require lot of custom code.

So far there might be few options:

1. Add support for most common npm registry, like `Verdaccio` (with help from community). Could have folder with resolvers.
2. Add support for self hosted index of plugins with predefined format, so by adding `URL` into Flipper settings we would allow internal plugin discovery and also future auto updating. This will require hosting some simple web app, and flipper could even include some simple package to just run this.

Let me know if any other ideas or approaches we could take to allow publishing Flipper plugins internally.

nikoant:
Hey @usrbowe! We have distribution mechanism for internal plugins which in addition to installation also supports auto-updates for plugins. I think it should be easy to adopt for option 2. I'll write up a doc this week on how the mechanism works and propose how we could extend it to support custom plugin indexes.

usrbowe:
Hi @nikoant,
maybe we could start by just sharing some high level overview of the the mechanism, and I could try something out first. Wha you think?

nikoant:
Hey @usrbowe, sorry for delay with that! Sounds good, yes. I opened a PR which moves rest of the internal plugin distribution code to open-source. Other code like UI for installing and updating plugins and dispatcher/reducer for firing downloads was already there.

nikoant:
I added some explanation of the current design and comments here: https://github.com/facebook/flipper/pull/3473. I'm happy to hear any suggestions on how we can re-use it for different orgs. 

Eventually I would also like to better unify "internal" distribution and "public" distribution via npmjs.org. For now they are two totally different branches which isn't very convenient.

usrbowe:
Nice, let me go through the PR first. Will try to came up with some proposal.

For the unification of public/internal plugins, you would like to use only npmjs.org? Because that would require to have a npmjs plan to publish private packages.

nikoant:
> For the unification of public/internal plugins, you would like to use only npmjs.org?

[...]

PR #3473, opened 2022-02-22
Make internal plugin distribution code publicly available

nikoant:
Summary:
This diff is the first one which addresses https://github.com/facebook/flipper/issues/3320.

In this diff we are making a part of the code used for internal Flipper plugin distribution in Meta also available publicly for re-using in other orgs.

Some explanation on how plugin installation and updates is designed now:
1) We periodically poll for plugins available for download. API for retrieving available plugins list is abstracted and will be different between public and fb versions, however all other logic is re-used.
2) In addition to "Enabled" and "Disabled" plugins in the left panel Flipper shows "Detected in App" list. Plugins in this list are those which are known compatible with the currently selected device/app, but not yet installed.
3) User can install any of "Detected in App" plugins by clicking to "Download and install" button near them in the left panel similarly to enabling plugins in "Disabled" list.
4) If we detect that for some installed plugin we have a newer version available for download - we download it silently and store on disk.
5) If the plugin for which we have new downloaded version is disabled - we update it silently without any notifications by loading new version from the disk and unloading the previous version from cache.
6) If the plugin for which we have new downloaded version is enabled then we avoid updating it automatically (because we need to reset plugin state in such case) and instead show notification on top of the plugin and ask user to reload it to apply new version. On reloading we reset the plugin state.
7) On Flipper startup we always update all plugins to their latest versions available on the disk.

Differential Revision: D34380308


desktop/flipper-ui-core/src/fb-stubs/pluginMarketplaceAPI.tsx
```diff
@@ -7,10 +7,10 @@
...
-}
+import {MarketplacePluginDetails} from '../reducers/plugins';
 
-export default () => {
-  // Marketplace is not implemented in public version of Flipper
-};
+export async function loadAvailablePlugins(): Promise<
+  MarketplacePluginDetails[]
+> {
+  return [];
```
nikoant, line 15: This is the place where we can hook returning details of the plugins available for downloading.

nikoant:
Other code related to plugin distribution was already on github before: 
- https://github.com/facebook/flipper/blob/main/desktop/flipper-ui-core/src/reducers/pluginDownloads.tsx
- https://github.com/facebook/flipper/blob/main/desktop/flipper-ui-core/src/dispatcher/pluginDownloads.tsx
- https://github.com/facebook/flipper/blob/main/desktop/flipper-server-core/src/plugins/PluginManager.tsx#L90 - downloads themselve. Expected plugin package format is tar.gz npm package.
- https://github.com/facebook/flipper/blob/main/desktop/flipper-ui-core/src/sandy-chrome/appinspect/PluginList.tsx#L279 - "Detected in App" plugin group where user can trigger downloading and installing a plugin.

usrbowe:
Tested this PR successfully with mocked data. 
So will try to create another PR to add option to provide "Plugin marketplace" registry as endpoint.

Found just one issue, when tried to download from 3rd party registry (verdaccio), that this code would throw error:

https://github.com/facebook/flipper/blob/main/desktop/flipper-server-core/src/plugins/PluginManager.tsx#L130-L134

Not sure if, this is really needed?

nikoant:
> Tested this PR successfully with mocked data. So will try to create another PR to add option to provide "Plugin marketplace" registry as endpoint.
> 
> Found just one issue, when tried to download from 3rd party registry ([verdaccio](http://verdaccio.org)), that this code would throw error:
> 
> https://github.com/facebook/flipper/blob/main/desktop/flipper-server-core/src/plugins/PluginManager.tsx#L130-L134
> 
> Not sure if, this is really needed?
> 
> Apart from this, it all works well. It shows discovered apps: image And also allow to download plugin and enable by default.

Great, thanks for checking that! We can remove check for content-type which throws or add an allowlist of content types there. What type Verdaccio returns?

usrbowe:
#### The `content-type` issue:
I just checked and it does report the `application/octet-stream`, but the string also includes the charset `application/octet-stream; charset=utf-8`.
So we could do something like this:
```js
// Util to parse content-type header
function parseHeaderValue(header) {
   const values = header.split(';');
  // remove white space
  return values.map(value => value.trim());
}

// Changes to: desktop/flipper-server-core/src/plugins/PluginManager.tsx@131
const parsedHeaders = parseHeaderValue(response.headers['content-type']);
if (!parsedHeaders.includes('application/octet-stream')) {
          throw new Error(
            `It looks like you are not on VPN/Lighthouse. Unexpected content type received: ${response.headers['content-type']}.`,
          );
        }
```

desktop/flipper-ui-core/src/dispatcher/pluginMarketplace.tsx
```diff
@@ -0,0 +1,209 @@
...
+          } else {
+            // If plugin was not installed before, then we should mark it as enabled
+            // to ensure it is automatically loaded after downloaded from Marketplace.
+            store.dispatch(setPluginEnabled(plugin.id, client.query.app));
+          }
+        }
+      }
+    }
+  }
+  if (isAutoUpdateDisabled() || !isLoggedIn().get()) {
```
usrbowe, line 180: This condition will never allow for auto update to happen, since the `isLoggedIn().get()` returns `false`.

So either we can rewrite the stub for OS, or remove those `isLoggedIn` conditions from OS code.

```js
const isLoggedInAtom = createState(false);
export function isLoggedIn(): Atom<boolean> {
  return isLoggedInAtom;
}
```

desktop/flipper-ui-core/src/dispatcher/pluginMarketplace.tsx
```diff
@@ -0,0 +1,209 @@
...
+    !getFlipperLib().isFB ||
+    getRenderHostInstance().GK('flipper_disable_plugin_auto_update') ||
+    getRenderHostInstance().serverConfig.env.FLIPPER_NO_PLUGIN_AUTO_UPDATE !==
+      undefined
+  );
+}
+
+function isPluginMarketplaceDisabled() {
+  return (
+    !getFlipperLib().isFB ||
```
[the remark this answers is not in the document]
usrbowe: yeah, for the `isPluginMarketplaceDisabled` and `isAutoUpdateDisabled` we can add new option in Flipper Settings.
We will need to provide url for the server with marketplace plugins and can also have option to control auto update (ON/OFF + interval)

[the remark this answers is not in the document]
nikoant: Agreed, sounds like a good set of options!

desktop/flipper-ui-core/src/dispatcher/pluginMarketplace.tsx
nikoant: Changing stub to always return true in OS sounds good to me

desktop/flipper-ui-core/src/dispatcher/pluginMarketplace.tsx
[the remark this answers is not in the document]
usrbowe: Alright, I've done some POC, so will create new PR on top of this to discuss the direction.

image

desktop/flipper-ui-core/src/fb-stubs/pluginMarketplaceAPI.tsx
```diff
@@ -7,10 +7,10 @@
...
-export async function loadPluginsFromMarketplace() {
-  // Marketplace is not implemented in public version of Flipper
-}
+import {MarketplacePluginDetails} from '../reducers/plugins';
 
-export default () => {
-  // Marketplace is not implemented in public version of Flipper
-};
+export async function loadAvailablePlugins(): Promise<
+  MarketplacePluginDetails[]
```
usrbowe, line 13: Would it be possible to provide a minimal schema of what is required fields for plugin to be detected and downloadable from the marketplace API?

Current typing also including fields for docs `publishedDocs`, which I believe is only useful for internal use in Meta/FB.

I would like to provide some basic schema, so others can create their own API easily to feed Flipper with plugins.

In general, the main object is basically:
```js
type MarketPlacePlugin = {
   ...package.json,
// flipper specific fields
specVersion: number; // not sure how to get this, since not present in package.json
availableVersions: [], // array of available plugins (type MarketPlacePlugin)
  downloadUrl: string; // provide a link to a tarball
  lastUpdated: Date; // need provide date from last published package
// those bellow, not sure if just keep as it is by default?
 isActivatable: false;
  isBundled: false;
  // Indicates whether plugin should be enabled by default for new users
  isEnabledByDefault: boolean;
  // Determines whether the plugin is a part of the Flipper JS bundle.
  isBundled: boolean;
  // Determines whether the plugin is physically available for activation in Flipper.
  isActivatable: boolean;
}
```

PR #3491, opened 2022-03-02
Enable Marketplace for Flipper

usrbowe:
## Summary

This is PR on top of: https://github.com/facebook/flipper/pull/3473

It adds an option to Settings to allow distribution of marketplace plugins. 

Also includes a simple fetch function to retrieve data from external API/server.

## Changelog

Allow marketplace plugins

## Test Plan

1. Enable marketplace
2. Provide custom marketplace server (it will serve the list of internal plugins with downloadURL)
3. Test if can see Available plugins and can download/remove the plugin
4. If new update for the plugin, it should also allow auto update


usrbowe:
@nikoant this is my few changes to allow the marketplace. Mainly its just the Settings and the API fetching.

For the Flipper, I think also need write a doc where we can explain what is the requirement for the server and possibly could provide some quick POC for lets say `express` server.

Further, I was thinking we could change settings modal a bit, so it has a tabs and we can hide the Marketplace options under some `Advance` tab, since it will not be used by all users.

nikoant:
That's super cool! Adding a separate tab for Settings modal is a great idea too as it becomes more and more bloated.

nikoant:
There are also few failing tests. Could you check them? You can run tests using "yarn test <path to test>".

usrbowe:
@nikoant yup, sounds good 👍 

I could share my existing `express` serve code, which I used internally to test Flipper marketplace and also planning to host internal plugins.

For the modal changes, maybe I can wait for this to be merged first and will create changes on top of this.

desktop/flipper-ui-core/src/dispatcher/pluginMarketplace.tsx
```diff
@@ -177,7 +184,8 @@ export function autoUpdatePlugins(
       }
     }
   }
-  if (isAutoUpdateDisabled() || !isLoggedIn().get()) {
+  // FIXME: this isLoggedIn blocks the logic
```
usrbowe, line 187: @nikoant forgot to remove this comment, as we discussed here: https://github.com/facebook/flipper/pull/3473#discussion_r814909991

So either I can push changes and remove the comment and change the stub value, or if already pulled to internal codebase, you could do the change there. Let me know which works better for you.

kenyee:
@usrbowe did you ever share your server code?

kenyee:
does this work w/ just a normal internal Artifactory server?

usrbowe:
Here is a simple demo server to get plugins from `Verdaccio.org` -> https://gist.github.com/usrbowe/f45c93384e57ea35054b0e7d41ee2557 

So likely for your use case, need rewrite the `npmAdapter` to fetch packages from Artifactory.

@kenyee

PR #3646, opened 2022-04-20
[Flipper-JS] fix: clear reconnect timeout on stop()

usrbowe:
## Summary

When using `flipperClient.stop()`, after `flipperClient.start()`, it's impossible to `start` new connection, since the previous instance would keep active `reconnect`, so would override the previous connection.

## Changelog

- clean reconnect timer when calling `flipperClient.stop()`

## Test Plan

1. `flipperClient.start('Demo app', { urlBase: 'null:8333' });` - this will fail connection, but keep retrying
2. `flipperClient.stop()` - should disconnect from WS and clear all reconnect timers
3. `flipperClient.start('Demo app', { urlBase: 'localhost:8333' });` - should connect succefully

### DIFF
--- a/js/js-flipper/src/client.ts
+++ b/js/js-flipper/src/client.ts
@@ -190,13 +190,13 @@ export class FlipperClient {
   }
 
   stop() {
-    if (!this.ws) {
-      return;
-    }
-
     if (this.reconnectionTimer) {
       clearTimeout(this.reconnectionTimer);
       this.reconnectionTimer = undefined;
+    }
+
+    if (!this.ws) {
+      return;
     }
 
     // TODO: Why is it not 1000 by default?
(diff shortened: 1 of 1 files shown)


js/js-flipper/src/client.ts
```diff
@@ -190,6 +190,11 @@ export class FlipperClient {
   }
 
   stop() {
+    if (this.reconnectionTimer) {
+      clearTimeout(this.reconnectionTimer);
+      this.reconnectionTimer = undefined;
+    }
+
     if (!this.ws) {
```
usrbowe, line 198: Note: Important to keep the clearTimeout above this condition, since the `reconnect` method sets the `this.ws = undefined`.

PR #3648, opened 2022-04-21
Add "MacOS" to supportedDevices

usrbowe:
This is already supported, when running Flipper on Mac, so just need update schema.

## Summary

One use case for `MacOS`, is to have a static plugin available to users before they launch any app/device. So this can be used as Welcome plugin, Documentation or any other "static" plugin not requiring any other device except the one running Flipper Desktop.

## Changelog

- add `MacOS` to supported devices

## Test Plan

- Script `lint` should pass.


usrbowe:
Ideally if we can make the `enum` more forgiving, since the `flipper-js` can also define it's own OS/Device type.

Issue #3751, opened 2022-05-25
Discussion: multiple host support in mocking


bangarharshit:
Currently, we provide a full URL for mock but a very common use case is to have multiple hosts (pops or data-centre). Support for mocking.

I was also looking at https://github.com/facebook/flipper/issues/2974 and https://github.com/facebook/flipper/issues/2077 and was thinking of a scaleable way to do it. There are a few ways to do it such as how Charles do it - [article](https://wilcovanes.ch/articles/mocking-api-requests-and-responses-through-charles-proxy/), and [article](https://www.thinktecture.com/en/tools/debugging-proxies-mocking-manipulating-api-charles-in-action/)

Happy to create a proposal and send PR

PR #3762, opened 2022-05-29
Prototype: Added support for multiple host in mock

bangarharshit:
## Summary
Apps usually communicate with multiple hosts to support pops and data centres. The problem currently is that for each host we need a different config in the mock tab in flipper. If the host change due to failover then the mocking stops working. Added support to enter the URL without a host in the flipper. We then check if the Url from the actual network request contains the mock URL without a host.

We can change the hint on flipper UI to suggest either entering the complete URL or without the host to allow matching across multiple hosts. 

Resolve:[3751](https://github.com/facebook/flipper/issues/3751)

Let me know if the approach looks good.

### DIFF
--- a/android/plugins/network/src/main/java/com/facebook/flipper/plugins/network/FlipperOkhttpInterceptor.java
+++ b/android/plugins/network/src/main/java/com/facebook/flipper/plugins/network/FlipperOkhttpInterceptor.java
@@ -23,6 +23,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


passy:
Nice! Thanks a lot for your contribution.

bangarharshit:
verified it by running locally.


do we need to change text here - https://github.com/facebook/flipper/blob/79023ee190d9e78a7c95bfb32912a4920fd6bfe4/desktop/plugins/public/network/request-mocking/MockResponseDetails.tsx#L146 to say that URL without host is supported as well.

PR #3825, opened 2022-06-20
Use supplied path when installing plugin from NPM to prevent escaping issues

### LINKED ISSUE #3819: Unable to install plugin with `/` in its name
## 🐛 Bug Report

When installing a plugin with `/` in the name (i.e. `@shopify/flipper-plugin-react-native-performance`), the installation fails with an error.

## To Reproduce

Using Flipper's UI, search for `@shopify/flipper-plugin-react-native-performance`, and click "Install". It will fail.

In the logs screen, there should be an error like:

```
Installation process of kind Install failed with: Error: ENOENT: no such file or directory, open '/path/to/tmp/@shopify__flipper-plugin-react-native-performance/package.json'
```

## Environment

macOS Monterey
Flipper 0.149.0 (and also `master` as of e43884cf11b2eb4029d2f0f13fa5ef2573a8ac17)

## Solution

The usage of `getPluginDirNameFromPackageName()` is inconsistent with how `live-plugin-manager` determines the directory.

In the above example, `live-package-manager` _does_ create the package at:

```
/path/to/tmp/@shopify/flipper-plugin-react-native-performance/package.json
```

But `getPluginDirNameFromPackageName()` is returning:

```
/path/to/tmp/@shopify__flipper-plugin-react-native-performance/package.json
```

So by patching [`installPluginFromNpm()`](https://github.com/facebook/flipper/blob/69bac4a3d6b5546ef6bd05b8124c33304380a758/desktop/plugin-lib/src/pluginInstaller.tsx#L99-L114) to set `pluginTempDir` as the result from `plugManNoDep.install()`, the plugin installed successfully.

For example:

```ts
export async function installPluginFromNpm(name: string) {
    // ...
    const result = await plugManNoDep.install(name);
    const pluginTempDir = result.location;
    return await installPluginFromTempDir(pluginTempDir);
    // ...
}
```

I suspect this has side effects for uninstalling/updating (or on other platforms?), as there are more references to `getPluginDirNameFromPackageName()` elsewhere in code. I'm happy to put up a naïve PR, but if you have a more context on the side effects, I'll leave it up to you.

Thanks!

notjosh:
## Summary

Resolves #3819

This PR fixes Flipper's plugin installation, so that scoped plugins (with `/` in the name) can be successfully installed (+removed).

In more detail, Flipper's function [`installPluginFromNpm()`](https://github.com/facebook/flipper/blob/69bac4a3d6b5546ef6bd05b8124c33304380a758/desktop/plugin-lib/src/pluginInstaller.tsx#L99-L114) relies on the `live-plugin-manager` package to handle the NPM package to a temporary directory.

Before this patch, Flipper would assume the name of the directory (incorrectly). This patch simply uses the directory as provided by `live-plugin-manager`'s result.

If we use `@shopify/flipper-plugin-react-native-performance` as a concrete example after this patch:

- `installPluginFromNpm()` is called to install `@shopify/flipper-plugin-react-native-performance`
- `plugManNoDep.install()` installs the plugin to `/path/to/tmp/@shopify/flipper-plugin-react-native-performance`
  - note: before this patch, it would have been assumed to be installed to: `/path/to/tmp/@shopify__flipper-plugin-react-native-performance` which does not exist
- `installPluginFromTempDir()` is called to install the package to `~/.flipper/installed-plugins/@shopify__flipper-plugin-react-native-performance/{version}` (where `{version}` is the current version, i.e. `1.0.0`)

Once the plugin is in `~/.flipper/installed-plugins/`, the escaping is correctly applied consistently by Flipper, and there are no further issues managing/using the plugin.

## Changelog

Fixed errors when installing scoped NPM plugins

## Test Plan

Ideally we'd have some unit tests, but mocking out the NPM fetching in `live-plugin-manager` seems like a bunch of work. If you have some shortcuts, let me know, because this would be useful to test across platforms.

But in the meantime, we can easily manually test this.

### Reproduce the failure

First, let's reproduce the failure:

- run Flipper 0.150.0 (or `yarn start` on master, etc)
- open plugin marketplace, attempt to install scoped plugins
- installation will fail (with ⚠️ icon) 
Screen Shot 2022-06-20 at 1 18 21 pm
- logs will show a failure to read `package.json` 
Screen Shot 2022-06-20 at 1 18 39 pm

### Demonstrate success

- run Flipper from this branch (via `yarn start`)
- open plugin marketplace, attempt to install scoped plugins
- installation will succeed, with notification to restart Flipper 
image
- (package can also be successfully used/upgraded/uninstalled)


baveku:
Does anyone check this PR?
This will help my flipper a lot.

lblasa:
> Does anyone check this PR? This will help my flipper a lot.

I'll take a look at this, sorry about the delay!

Issue #3820, opened 2022-06-15
Can't zoom flipper desktop app


Flam3rboy:
## 🐛 Bug Report
<img width="351" alt="image" src="https://user-images.githubusercontent.com/34555296/173853820-9782628b-b097-4ce6-b6ce-0f6746426dca.png">

```js
Uncaught Exception:
TypeError: Cannot read properties of undefined (reading 'setZoomFactor')
at click (/Applications/Flipper.app/Contents/Resources/app.asar/main.bundle.js:7:2194)
at MenuItem.click (node:electron/js2c/browser_init:73:1908)
at EventEmitter.a._executeCommand (node:electron/js2c/browser_init:81:2453)
```

## To Reproduce
Install the latest flipper desktop app and try to increase the zoom level by click on the `top bar -> View -> Zoom in`

## Environment
Flipper Desktop Version 0.149.0 (50.0.0)
macOS Monterey 12.4 (21F79)

see-mik:
Hello.
The issue is still reproducing. 
Flipper (0.156.0)
macOS (12.4)

mweststrate:
Closing as #3840 will be part of next weeks release. Thanks for reporting and apologies for late follow up!

PR #3840, opened 2022-06-20
Fixed reload & zoom in Flipper Desktop application.

### LINKED ISSUE #3839: Exception while trying to reload after Electron upgrade
## 🐛 Bug Report
After the upgrade to Electron in `92cdb810969480c892b906724b7db495bd9b4d0a` (https://github.com/facebook/flipper/commit/92cdb810969480c892b906724b7db495bd9b4d0a), the reload functionality no longer works. Reloading the application results in the following exception:
exception

## To Reproduce
1. Open Flipper.
2. From the Menu Bar: View -> Reload (or Cmd+R).

## Environment
Flipper Desktop v.0.150.0

mhrheaume:
## Summary

After the upgrade to Electron in `92cdb81`, the reload functionality no
longer works. I believe this is due to the main menu creation being
moved to the main process where the `window` global variable is undefined
(whereas previously it was on the render process, where `window` would
be defined). Rather than rely on the `window` global, when a reload is
requested just reload the given `focusedWindow`.

Similarly, zooming was also broken, so fix that by going through `focusedWindow` rather than `webFrame` (since `webFrame` only exists in the render process).

Fixes #3839, #3820

## Changelog

Fixed reload & zoom in Flipper Desktop application.

## Test Plan

Built the application locally, verified that reload worked as expected (and would reload the application), verified that zooming (in, out, actual size) worked as expected.


Flam3rboy:
When will this get merged?

mweststrate:
Sorry for the late follow up! Merging now, should be part of next weeks release.

mweststrate:
See also #3987 for test plan

PR #3841, opened 2022-06-21
docs: remove duplicated section for react-native setup

jeremybarbet:
## Summary

- Section with same explanation defined twice

## Changelog

- docs: remove duplicated section for react-native setup

## Test Plan

- Check the react-native setup page


mweststrate:
Thanks for spotting! Closing as it is a duplicate of #3726. Thanks for taking the effort nonetheless!

PR #3932, opened 2022-07-23
Fix incorrect getting started link in README.md

nikit19:
## Summary

Add the correct getting started link in README.md


mweststrate:
Thanks for spotting!

mweststrate:
similar diff already landed

Issue #3805, opened 2022-06-10
Getting white background while hovering over an object (dark  mode)


ZeeshanAhmadKhalil:
## 🐛 Bug Report

When hovering over any object in the response to network request or in the redux state, got white background which hides out all keys in the object since keys and background have the same colour. Please take a look at the following screenshot for a better understanding.

![image](https://user-images.githubusercontent.com/41861952/173108647-fdfffd0a-000a-4f5c-ba04-443122f68a8b.png)

## To Reproduce

Download the latest version of the flipper `0.149.0` and switch to the dark mode.

## Environment

Windows 10 OS
flipper 0.149.0

ZeeshanAhmadKhalil:
I should use the older version for now. Don't know what to do if this persists in the upcoming versions as well.

dennisdeng2002:
Agreed, quite annoying as dark mode user

ZeeshanAhmadKhalil:
Still not resolved in `0.150.0`

plwai:
Seems like dark mode is not handled in the `v0.149.0` hover feature.

https://github.com/facebook/flipper/blob/40e65901bd6191603248a0ec2b3d31b13ed42e26/desktop/flipper-plugin/src/ui/data-inspector/DataInspectorNode.tsx#L53

ZeeshanAhmadKhalil:
Resolved in `0.158.0`

PR #3967, opened 2022-08-03
fix data inspector white hover background in dark mode

timostroehlein:
## Summary

Fixes the white color background in dark mode for the data-inspector plugin (regarding issue #3805). Now instead of a hardcoded color, the selectionBackgroundColor from theme will be used.

## Changelog

- fix data inspector plugin white hover background color in dark mode

## Test Plan

Previous:
image

Dark mode:
image

Light mode:
image


samholmes:
This looks like the correct solution to me.

mweststrate:
This was just fixed last week, and should have been fixed in todays release. Thanks for spotting and fixing though anyways!

Issue #3947, opened 2022-07-28
Discussion: Feature request - add a config flag to disable the "new update available" check


kenyee:
Use case: updates to Flipper sometimes break plugins.
We'd like to lock our devs' version of flipper and the plugins to a specific version.

Please add a configurable (via some text file in a known location) flag that would let us disable this update check.

rasaha91:
For my organization's use case, it might be preferable to expose this as a command line flag (e.g. `--no-update-notifications`).

kenyee:
Was looking at the .flipper/config.json file and noticed an "updaterEnabled" flag and got excited:
```
{"pluginPaths":[],"disabledPlugins":[],"darkMode":"light","updaterEnabled":true,"launcherEnabled":true,"lastWindowPosition":{"x":328,"y":38,"width":1400,"height":994}}
```

Only to find that nothing in the code does anything w/ this flag...searching github only shows this file using "updaterEnabled":
https://github.com/facebook/flipper/blob/b4fc108c2e303deb4488b07a4332fa916c85ddc4/desktop/static/setup.tsx

kenyee:
Tried setting this updaterEnabled flag to false when starting Flipper 0.156.0 and it autoprompted anyways :-(

Seems like the updater plugin doesn't read this config flag like it should...

kenyee:
was going to have a go at trying to fix this, but the Flipper desktop app doesn't seem to run when I check out this repo and follow the directions: https://github.com/facebook/flipper/issues/3988

kenyee:
hmm...there's also a --no-updater flag, but it's probably for the Electron framework: https://www.electronjs.org/docs/latest/tutorial/updates

kenyee:
Going to add a "notifyUpdateAvailable" flag for config.json.  The existing updaterEnabled flag might have been used for autoupdates?  No code seems to use it though.
When configuring the config.json, it's probably safer to set both these flags to false in case the updaterEnabled flag ever gets enabled.

kenyee:
renamed flag to suppressPluginUpdateNotifications FYI

aigoncharov:
Should be resolved by https://github.com/facebook/flipper/pull/3992

PR #3992, opened 2022-08-11
Adds a notifyAvailableUpdate flag to config.json to disable update

kenyee:
## Summary

We would like to version control Flipper and some of our custom plugins that are installed on developers' systems.
Flipper by default prompts users to upgrade so they sometimes do the update and then all our custom plugins break because they were compiled for an older version.
See https://github.com/facebook/flipper/issues/3947 for feature request info.

## Changelog

Adds notifyAvailable flag to config.json to disable prompting for users that "an update is available"

## Test Plan

Tested by running locally.
Had to comment out the isProduction() check to confirm this it worked properly because this flag is false on dev versions.
Couldn't figure out how to manually test the handleOpenPluginDeeplink.tsx change but made a similar change there; happy to test that if you can tell me how to exercise that path.


desktop/flipper-common/src/settings.tsx
```diff
@@ -76,7 +76,7 @@ export type LauncherSettings = {
   ignoreLocalPin: boolean;
 };
 
-// Settings that primarily only apply to Eelectron atm
+// Settings that primarily only apply to Electron atm
```
kenyee, line 79: fixed a few typos I noticed when looking at the code

desktop/static/main.tsx
```diff
@@ -216,7 +216,7 @@ app.on('ready', async () => {
               {allowFileAccess: true},
             );
           } catch (e) {
-            console.error('Failed to loa React devtools from disk: ', e);
+            console.error('Failed to load React devtools from disk: ', e);
```
kenyee, line 219: fixed typo

desktop/flipper-common/src/settings.tsx
```diff
@@ -90,6 +90,8 @@ export type ProcessConfig = {
   launcherMsg: string | null;
   // Controls whether to delegate to the launcher if present.
   launcherEnabled: boolean;
+  // Control whether to enable "update available" notifications
+  notifyAvailableUpdate: boolean;
```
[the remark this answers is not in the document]
mweststrate: Would you mind negating the flag, so that the behavior is more intuitive if not set, in which case it is the same as `false`?. E.g. `suppressPluginUpdateNotifications`?

desktop/static/main.tsx
mweststrate: thanks for spotting!

kenyee:
> For testing deeplinks: Gear icon > Plugin developers > Trigger deeplink

Thanks...this seems to exercise that codepath:
flipper://open-plugin?plugin-id=Example

desktop/static/setup.tsx
```diff
@@ -33,6 +34,7 @@ export type Config = {
   updaterEnabled?: boolean;
```
rasaha91, line 34: Curious, what does this config setting do then?

desktop/static/setup.tsx
```diff
@@ -83,6 +85,10 @@ export default async function setup(argv: any) {
     updaterEnabled: argv.updater,
     launcherEnabled: argv.launcher,
     launcherMsg: argv.launcherMsg,
+    suppressPluginUpdateNotifications:
```
rasaha91, line 88: Is there documentation that needs to be updated as well? I'm not sure how end users are supposed to know about config settings for flipper.

desktop/static/setup.tsx
kenyee: FWIW, I never found any usages of this in the code...initially thought this was used to disable the updater when I was trying to see if there was an undocumented feature...

desktop/flipper-common/src/settings.tsx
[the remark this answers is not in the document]
kenyee: @mweststrate changes made...anything else I should change?

desktop/static/setup.tsx
kenyee: there are no docs for the config.json file AFAIK...would be nice if there were but that's probably a separate PR...

PR #4087, opened 2022-09-13
Update Podfile setup

ray-holland-es:
`use_flipper` is deprecated, updated the Podfile setup to match the docs for Automatic setup

## Summary

The docs for setting up Podfile are out of date and using a method that is deprecated.

## Changelog

Update React Native - Manual iOS Setup Documentation

## Test Plan

n/a


aigoncharov:
@ray-holland-es should it be a separate section for React Native 0.69+?

PR #4204, opened 2022-10-12
Add example for marketplace server

usrbowe:
## Summary

Continue on top of this change: https://github.com/facebook/flipper/pull/3491

Add a server to host a list of plugins for Flipper Marketplace.

This code only includes very basic `express` server, with a example `npmAdapter` to demonstrate how to generate the endpoint list of plugins from a specific NPM server (Verdaccio).

The code does not cover support for any other NPM server implementations. 

Likely we could include demo for private NPM account.

## Test Plan

Test if the output will successfully show in the Flipper App under Discovered tab.


aigoncharov:
@usrbowe great job! Do you think you could create a dedicated repo for it and add a link to your repo to Flipper's documentation instead? I am afraid we won't be able to properly support it here as we do not use it internally

Issue #4182, opened 2022-10-05
Release action is failing to upload artifacts


rasaha91:
## 🐛 Bug Report
See https://github.com/facebook/flipper/actions/runs/3150989565/jobs/5124580334 where we are having a failure to upload release artifacts AND failing to open an issue for this failure, which is why there is no issue tracking it.

## To Reproduce

N/A

## Environment

N/A

rasaha91:
I debugged this issue and I believe the problem is the fact that https://github.com/passy/github-upload-release-artifacts-action uses a docker image based on Alpine linux, which uses a minimal version of libc. The error messages in the failed android publishing and release workflows is that ghr cannot be found, despite downloading it to /usr/local/bin. I verified that /usr/local/bin/ghr does exist. I ran ldd to see what the issue was (see https://github.com/rasaha91/flipper/actions/runs/3238534082/jobs/5306864002#step:2:112). This is effectively the same issue in this SO thread: https://stackoverflow.com/questions/67672392/docker-image-with-alpine-linux-an-executable-file-is-definitely-there-but-canno

I don't know why this would be happening all of a sudden as the dockerfile for the action hasn't been updated since November 2021. 

@passy and @lblasa in case you have ideas.

rasaha91:
Upon further inspection, we download the latest release of ghr. This release: https://github.com/tcnksm/ghr/releases/tag/v0.16.0 was released 17 days ago, and the release action started failing around 14 days ago, so the timeline seems right. My guess is something breaking on Alpine linux was introduced in that release. Perhaps a simple fix would be to hardcode the release tag to download.

rasaha91:
@aigoncharov hoping to get some traction on this issue.

rasaha91:
I verified that using v15.0 of ghr works. That version is statically linked, which is why we don't have any problems.

rasaha91:
PR for the fix is above. @passy  can you please merge the PR and create a new release?

rasaha91:
Upstream PR is here: https://github.com/Roang-zero1/github-upload-release-artifacts-action/pull/5

aigoncharov:
@rasaha91 thanks for reporting the issue and coming up with a fix! @passy is on a leave though. I've [forked your fork](https://github.com/aigoncharov/github-upload-release-artifacts-action) (xD) to get unblocked.
Many thanks for making Flipper better!

aigoncharov:
[...]

PR #4209, opened 2022-10-12
Fix inability to create issue upon Publish failure

rasaha91:
## Summary

The publish job in the release workflow has been failing, but an issue has failed to be created due to the template file not being found. As per the error message, the template file isn't found because we need to check out the repo first.

## Changelog

Fix problem with auto-creating issue upon publish failure.

## Test Plan

Not able to run this locally, so we will need to wait upon the next release to ensure the issue gets created.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -181,6 +181,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


rasaha91:
https://github.com/facebook/flipper/issues/4182

PR #4220, opened 2022-10-17
Update action in release workflow

rasaha91:
## Summary

The release workflow is using a version of the github release action that is broken, due to ghr now being dynamically linked and not compatible with Alpine Linux's minimal version of libc. The fix is to use the version of the action which installs gcompat so that ghr can run on the docker image.

## Changelog

Update release workflow to use version of action compatible with latest version of ghr.

## Test Plan

N/A


rasaha91:
@aigoncharov as FYI, this was a missed replacement.

PR #4275, opened 2022-10-28
Fix searching for FlipperClient.h headers in productions builds if PRODUCTION=1 passed

MuhmdRaouf:
## Summary
In newer builds of react-native if you passed parameters PRODUCTION=1, flipper pods won't get installed
you can see that in their codebase here [react_native_pods.rb](https://github.com/facebook/react-native/blob/6a43fafd78d581f63c664b9af6d10828ac7f77fa/scripts/react_native_pods.rb#L140)

so when building ios `react-native-flipper` set `compiler_flags = '-DFB_SONARKIT_ENABLED=1'` in the podspec file
which initializeFlipper and lead to the following error: 
`node_modules/react-native-flipper/ios/FlipperReactNativeJavaScriptPluginManager.h:12:9: 'FlipperKit/FlipperClient.h' file not found` 

## Changelog
use same condition as react-native to wrap `s.compiler_flags = compiler_flags`
and conditionally pass `compiler_flags = '-DFB_SONARKIT_ENABLED=1'` to compiler_flags to avoid build problems 

## Test Plan
To reproduce
- create new react-native app
- install react-native-flipper `yarn add react-native-flipper`
- run `env PRODUCTION=1  npx pod-install `
- run `yarn ios`

Before patch:
```sh
node_modules/react-native-flipper/ios/FlipperReactNativeJavaScriptPluginManager.h:12:9: 'FlipperKit/FlipperClient.h' file not found
```

After patch:
```sh
▸ Build Succeeded
success Successfully built the app
```

### DIFF
--- a/react-native/react-native-flipper/react-native-flipper.podspec
+++ b/react-native/react-native-flipper/react-native-flipper.podspec
@@ -25,6 +25,10 @@ Pod::Spec.new do |s|
   s.source_files = "ios/**/*.{h,m,swift}"
   s.pod_target_xcconfig = { "HEADER_SEARCH_PATHS" => "\"${PODS_ROOT}/Headers/Public/FlipperKit\"" }
   s.requires_arc = true
-  s.compiler_flags = compiler_flags
+  if ENV['PRODUCTION'] == '1'
+    Pod::UI.puts "#{s.name}: Found PRODUCTION=1, #{s.name} will be disabled for production builds"
+  else
+    s.compiler_flags = compiler_flags
+  end
   s.dependency "React-Core"
 end


mweststrate (changes requested):
We deliberately do *not* want to be included in production builds, as it is a huge risk to both security and performance. If you want to include Flipper in prod builds at your own risk, feel free, but we don't want to provide any infra to encourage such thing as we consider it a really bad idea.

Let me know if I misunderstood the PR :)

MuhmdRaouf:
@mweststrate Hey, this PR was focused on fixing the problem located here

https://github.com/facebook/flipper/blob/1adcf2bc68acf137c41d99c967012b5dfc80dff5/react-native/react-native-flipper/ios/FlipperReactNativeJavaScriptPlugin.m#L8

as you can see you check if `defined(DEBUG) || defined(FB_SONARKIT_ENABLED)` 
and defined(FB_SONARKIT_ENABLED) is always true due PodSpec file here 

https://github.com/facebook/flipper/blob/1adcf2bc68acf137c41d99c967012b5dfc80dff5/react-native/react-native-flipper/react-native-flipper.podspec#L9

which leads Flipper to work even in a production env because of the OR condition

I hope that makes sense so far.

so to this PR, If I make  `env PRODUCTION=1 npx pod-install`

react-native will exclude Flipper pods from installing and when making iOS build the condition `defined(FB_SONARKIT_ENABLED)` will make the lib runs to look for `FlipperClient.h` which will cause the error 

`node_modules/react-native-flipper/ios/FlipperReactNativeJavaScriptPluginManager.h:12:9: 'FlipperKit/FlipperClient.h' file not found` cause flipper pods haven't got installed.

so this PR prevents this error from happing as it's removing the ` compiler_flags = '-DFB_SONARKIT_ENABLED=1' `
which will lead both conditions to false and build complete successfully.

also, I would suggest changing the condition here 

https://github.com/facebook/flipper/blob/1adcf2bc68acf137c41d99c967012b5dfc80dff5/react-native/react-native-flipper/ios/FlipperReactNativeJavaScriptPlugin.m#L8
 to be nested instead 

 ```swift
 #if defined(DEBUG)
 #if defined(FB_SONARKIT_ENABLED)
  // logic goes here
#endif // FB_SONARKIT_ENABLED
#endif // DEBUG
 ```
 I would love to make that PR too, just let me know how it will goes with this PR and we can go from there.

mweststrate:
@lblasa would you mind taking a closer look on the above? thanks!

robinclaes:
When setting `:flipper_configuration => FlipperConfiguration.disabled` in the Podfile, the React-native build indeed fails with  
`node_modules/react-native-flipper/ios/FlipperReactNativeJavaScriptPluginManager.h:12:9: 'FlipperKit/FlipperClient.h' file not found`

I think this error should not occur when disabling Flipper, regardless of the environment variable (`PRODUCTION`) being set or not.
So I think the solution in the PR might not be the ideal solution.

PR #4277, opened 2022-10-29
Pipe through --updater/--no-updater flag for flipper.exe

rasaha91:
## Summary

In our organization, Flipper is distributed in a version controlled way. As a result, we do not want users to manually update or receive prompts to update when a new version is available. There is already a `--updater` flag in the command line arguments for flipper.exe, but it isn't piped through. This change pipes it through and disables all update related UI when `--no-updater` is passed in.
## Changelog

Support --updater and --no-updater options for flipper.exe

## Test Plan

Ran `yarn build --win` in flipper/desktop, and launched flipper.exe from flipper/dist/win-unpacked with the `--updater`, `--no-updater` and no flags and ensured the proper behavior was observed (update UI shows by default or when `--updater` is specified, and doesn't show when `--no-updater` is specified).

### DIFF
--- a/desktop/flipper-common/src/settings.tsx
+++ b/desktop/flipper-common/src/settings.tsx
@@ -77,7 +77,7 @@ export type LauncherSettings = {
 };
 
 // Settings that primarily only apply to Electron atm
-// TODO: further separte between flipper-ui config and Electron config
+// TODO: further separate between flipper-ui config and Electron config
 export type ProcessConfig = {
   disabledPlugins: string[];
   lastWindowPosition: {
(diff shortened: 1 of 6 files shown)


rasaha91:
@aigoncharov could you take a look at this PR and import it/provide feedback?

rasaha91:
Ping. @aigoncharov @passy or @nikoant could one of you please take a look at this PR?

passy:
Thanks for preparing this, @rasaha91! One thing that isn't quite clear to me from looking at the code now: Does this change the default behaviour at all given that it's not initialised? We'd quite like to retain the updater behaviour being enabled by default. 

Rest looks good to me!

passy:
Thanks for fixing this. That should have already been the case and probably broke as part of some refactoring.

PR #4366, opened 2022-12-07
Fixed a bug where the screenshot of some Android phones was unavailable

kongxiaojun:
## Summary

The 'screenrecord' command is not in some Android phones e.g OPPO. So 'screenrecord' command can't be used as a basis for determining whether or not you can take a screenshot. Replace it with 'screencap' command  to determine whether you can take a screenshot.

## Changelog

Fixed a bug where the screenshot of some Android phones was unavailable

## Test Plan

It passed the test on my Mac and OPPO phone Reno2.


passy:
Oh, that's a great catch. I didn't know these differences existed.

Issue #3799, opened 2022-06-06
Crash when trying to remove a mock network call


davis-isaac:
## 🐛 Bug Report
Attempting to remove a mock network call causes a crash

> PLUGIN "NETWORK" ENCOUNTERED AN ERROR DURING RENDER
> TypeError: Cannot read property 'requestUrl' of undefined
>     at file:///private/var/folders/gp/wsnqxfjs64g60x7gcdbtbwqr0000gn/T/AppTranslocation/BA9BC7E5-5A2A-429B-87BE-A92E2741810B/d/Flipper.app/Contents/Resources/app.asar/bundle.js:5706:1452
>     at Array.reduce (<anonymous>)
>     at _duplicateIds (file:///private/var/folders/gp/wsnqxfjs64g60x7gcdbtbwqr0000gn/T/AppTranslocation/BA9BC7E5-5A2A-429B-87BE-A92E2741810B/d/Flipper.app/Contents/Resources/app.asar/bundle.js:5706:1395)
>     at file:///private/var/folders/gp/wsnqxfjs64g60x7gcdbtbwqr0000gn/T/AppTranslocation/BA9BC7E5-5A2A-429B-87BE-A92E2741810B/d/Flipper.app/Contents/Resources/app.asar/bundle.js:5706:2395
>     at Object.Ji [as useMemo] (file:///private/var/folders/gp/wsnqxfjs64g60x7gcdbtbwqr0000gn/T/AppTranslocation/BA9BC7E5-5A2A-429B-87BE-A92E2741810B/d/Flipper.app/Contents/Resources/app.asar/bundle.js:138:70803)
>     at _e.useMemo (file:///private/var/folders/gp/wsnqxfjs64g60x7gcdbtbwqr0000gn/T/AppTranslocation/BA9BC7E5-5A2A-429B-87BE-A92E2741810B/d/Flipper.app/Contents/Resources/app.asar/bundle.js:61:6829)
>     at ManageMockResponsePanel (file:///private/var/folders/gp/wsnqxfjs64g60x7gcdbtbwqr0000gn/T/AppTranslocation/BA9BC7E5-5A2A-429B-87BE-A92E2741810B/d/Flipper.app/Contents/Resources/app.asar/bundle.js:5706:2389)
[...]

PR #4403, opened 2022-12-21
Fix a crash when trying to remove a mock network call

kongxiaojun:
## Summary

Crash when trying to remove a mock network call

Issue: https://github.com/facebook/flipper/issues/3799

## Changelog

Read mock routes type error.Transform it from array to object.

## Test Plan

Test passed on my Mac.


passy:
Thanks!

PR #4506, opened 2023-02-08
Fix: WS should not reconnect on 1001

usrbowe:
Due to bug in ReactNative `WebSocket` implementation, the `ws.close(1000)` method will not carry the provided `onclose` event number. It would close connection with `1001` instead.

Technically, the `1001` also seems as normal status code to close WS connection, so likely we don't wanna use reconnect anyway.

This only impacts older RN versions (iOS), as it was later fixed: https://github.com/facebook/react-native/pull/24950/files

## Changelog

- Do not reconnect when WS close code is `1001` - GoingAway

## Test Plan

Run `flipperClient.stop()`, then the connection should not be reconnected (only reproducible on RN env.)

### DIFF
--- a/js/js-flipper/src/client.ts
+++ b/js/js-flipper/src/client.ts
@@ -237,7 +237,7 @@ export class FlipperClient {
     };
     this.ws.onclose = ({code}) => {
       // Some WS implementations do not properly set `wasClean`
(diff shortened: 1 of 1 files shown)


passy:
Great catch! Thanks a lot!

PR #4566, opened 2023-03-03
Fix example code on setup leakCanary plugin

Vova-SH:
## Summary

Fix sample for install leak canary plugin

## Changelog

add brackets on sample code

## Test Plan

-


passy:
Thank you!

PR #4734, opened 2023-05-10
update documentation for using react-native-flipper > 0.182.0

mark-zacharias:
## Summary
The current documentation is missing a vital step to using flipper greater than 0.182.0.  This small change should save a lot of time for folks looking to upgrade.
Since fresco is longer packaged with flipper the guidance to upgrade along with the react-native default code for the build.gradle file needs to be updated

## Changelog

update documentation for using react-native-flipper > 0.182.0

## Test Plan

follow current documentation to upgrade with a fresh react-native, note the build failure.
Apply the change outlined in the proposed docs and enjoy the happy.

### DIFF
--- a/docs/getting-started/react-native.mdx
+++ b/docs/getting-started/react-native.mdx
@@ -37,6 +37,10 @@ Android:
 1. Bump the `FLIPPER_VERSION` variable in `android/gradle.properties`, for example: `FLIPPER_VERSION=0.190.0`.
 2. Run `./gradlew clean` in the `android` directory.
 
+For react-native-flipper > 0.182.0 also modify the `app/build.gradle` file to lock the fresco version to 0.182.0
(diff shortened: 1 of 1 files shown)


DomiR:
While you are at it, maybe fix the example app [here](https://github.com/facebook/flipper/blob/cd67ce59f7dd70cbf459f0d4698071af6ee0abc8/react-native/ReactNativeFlipperExample/android/app/build.gradle#L210) as well...

passy:
This is not what we recommend. The plugin is now published under the Fresco org. Instead of hard-coding an old version that will never receive any updates, you should instead depend on:

```
implementation group: 'com.facebook.fresco', name: 'flipper-fresco-plugin', version: '3.0.0'
```

passy (changes requested):
Left a comment on the thread.

PR #4811, opened 2023-06-12
fix: prefixed plugin package name

usrbowe:
## Summary

When distributing plugin over Marketplace. Currently there is an issue with downloading a plugin to `tmp` folder if the plugin `packageName` starts with `@company-prefix/flipper-plugin-name`. It would throw `ENONET` error while trying to stream the response data into `tmp` folder.

## Changelog

- add support for prefixed plugin package names (`@shopify/flipper-plugin-name`,...)

## Test Plan

- Try install a plugin from Marketplace which has `/` in the plugin `package.json` `name`.


passy:
Thanks for catching that! I think we've just missed the case where you could have prefixed package names.

PR #4899, opened 2023-06-29
[Cleanup] - Delete unnecessary lock file

Pranav-yadav:
## Summary

Delete unnecessary lock file: `./react-native/yarn.lock` which was checked out but isn't necessary.
And, `gitignore` it.

## Changelog

[General] [Removed] - Delete unnecessary lock file.

## Test Plan

- None


passy (changes requested):
Sorry, but we do need these. It's really important for us given the sheer number of dependencies and transitive dependencies, that we know that the currently checked-in configuration works. If we remove the locks, then virtually every single user will get a different set of dependencies given the pace of the ecosystem. So we'd rather keep them and manually bump the dependencies.

Pranav-yadav:
>we do need these. It's really important for us given the sheer number of dependencies and transitive dependencies, that we know that the currently checked-in configuration works.

@passy I may be unaware of the full context here, but could provide more context on how **_empty lockfiles_** are ensuring a set of specific dependencies and their versions?
Those lock files are **empty**, they don't even have a single entry of digest for any package.

>So we'd rather keep them and manually bump the dependencies.

Can you provide more context here as well, like which dependencies, these lock files don't have adjacent `package.json` or equivalent pkg metadata files?

P.S.: The thought behind this PR was that these lock files were checked out by mistake :)

passy:
I see! That's actually a good question. If they don't get regenerated if we run `yarn` again, then they should probably go away indeed.

.gitignore
```diff
@@ -14,6 +14,12 @@ manifest.json
 # conflicts with FB internal infra
 .watchmanconfig
 
+# Don't checkout the unnecessary yarn/npm lock files
+yarn.lock
```
passy, line 18: Ah, I think this is what confused me. Could you specialise this to only refer to `./yarn.lock`? Then we don't accidentally exclude future real lock files that we DO want to check in. :)

passy (changes requested):

.gitignore
Pranav-yadav: ahhh. I remember doing that only initially (relative path), but for some reason vscode did NOT *un-highlight* the lock file (suggesting us that it's ignored by git) so added direct path and it worked, that could be due to some issue with vscode's internal git plugin, since I knew `.gitignore` supports (prefers) relative paths.

Thanks. Pushing now...

PR #4900, opened 2023-06-29
Refactor: Use React & ReactDOM `18.x` features and methods in `flipper-ui-core`

Pranav-yadav:
## Summary

This diff refactors `flipper-ui-core` to use React and ReactDOM `18` features and methods.
It's already using React & ReactDOM `18.x` *experimental* version.

Such as using,
- `root.render(...)` instead of `ReactDOM.render(...)`
- named import for `createPortal`

Also, formats most of the files to keep "import react" statements at the top.

## Changelog

[Internal] [Changed] - Refactor: Use React & ReactDOM `18.x` features and methods in `flipper-ui-core`

## Test Plan

- Tests and builds should pass.

### DIFF
--- a/desktop/flipper-ui-core/src/chrome/ChangelogSheet.tsx
+++ b/desktop/flipper-ui-core/src/chrome/ChangelogSheet.tsx
@@ -7,8 +7,8 @@
  * @format
  */
 
-import {Markdown} from '../ui';
 import React, {Component} from 'react';
+import {Markdown} from '../ui';
 import {reportUsage} from 'flipper-common';
 import {Modal} from 'antd';
 import {Dialog, theme} from 'flipper-plugin';
--- a/desktop/flipper-ui-core/src/chrome/ConsoleLogs.tsx
+++ b/desktop/flipper-ui-core/src/chrome/ConsoleLogs.tsx
@@ -7,8 +7,7 @@
  * @format
  */
 
-import {useMemo} from 'react';
-import React from 'react';
+import React, {useMemo} from 'react';
 import {Console} from 'console-feed';
 import type {Methods} from 'console-feed/lib/definitions/Methods';
 import type {Styles} from 'console-feed/lib/definitions/Styles';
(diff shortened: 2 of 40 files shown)


Pranav-yadav:
>**Note**
The conflict is due to a **temporary** change in `main` 2b17ac76de6973d99cb6e38835c8acb7748e544b - which will be hopefully reverted on Monday.

Pranav-yadav:
@lblasa Can I know why your commit (https://github.com/facebook/flipper/commit/9810edcdfbf0c8e9f096ddd102a89c886ec582b7) supersedes this PR? I assume you didn't know about this PR; right?
It doesn't matter now, but I wanted to put it forward... cc: @passy 

Also, I assume everyone in OSS follows "Check Before if it exists or done already" when it comes to pushing code and opening issues.

I submitted this PR on "June 29" (your commit "July 6") and since then *this PR was "not reviewed" or even no one commented about when it'll be reviewed*.

@passy One thing I try to avoid while contributing to OSS is **"wastage of time and duplication of efforts"** 😞.

Lmk, if these changes are good to ship, so I can rebase.

lblasa:
Hi @Pranav-yadav so sorry I missed this PR! TBH, this change looks way better than the one I have.

That change I submitted was ultimately reverted as it was causing some issues. Mainly, we need to ensure React 18 is used throughout because of this: 

https://github.com/react-component/trigger/issues/288

One of our colleagues look into it and there was a few dependencies that needed updating including antd. 

The issues we found were some unresponsiveness because of unnecessary renders. From his notes:

Tried bumping both rc-align and rc-trigger and it definitely changes things, less permanent freezes but still seeing some issues, sometimes tooltips don't appear or are very delayed. I'm also seeing a component take 10 seconds to render due to layout effects.   Overall it seems that antd (at least the version we have doesn't support react 18 well yet).   I think the next step might be to try to bump antd but that is a whole other story due to their breaking changes.

Pranav-yadav:
No worries, Lorenzo. 😃 I understand it's difficult to keep track of things in OSS world.

That's an interesting issue. Old libraries/packages do have some issues w/ React 18.
Yup. I recently did migrations to React 18 in other private projects and I can tell you it's very difficult to get everything without breaking changes or major refactors.

I can't promise but will look into it and will let u know. 👍

Issue #1480, opened 2020-08-18
Android Sample App Post Request failing with 404 error


bizzguy:
## 🐛 Bug Report

The Android Sample app allows user to "Send POST Request" which can be used to demonstrate the Network Plugin.  However, the request is no longer working.  It fails with a 404 (Page Not Found) error.  It appears the the service used to host the request (mockable.io) is no longer configured correctly.

## To Reproduce

1) Start Flipper
2) Start the Android Sample App
3) Issue the post request from the Sample App

![image](https://user-images.githubusercontent.com/337874/90540929-81831300-e147-11ea-8958-50d99585f330.png)

4) The request never completes in Flipper

![image](https://user-images.githubusercontent.com/337874/90541042-aa0b0d00-e147-11ea-9cfa-fe73c5aca8be.png)

Note:  Here is the code that makes the request:

![image](https://user-images.githubusercontent.com/337874/90541074-b7c09280-e147-11ea-8d19-87f33d14322b.png)

## Environment

Flipper 0.52.1

bizzguy:
@mweststrate 
Could you add the "plugin-network" label to this issue?  I could not figure out how to do it myself.  I'm assuming that I don't have access.  

Thanks.

mweststrate:
Done. P.s. I think it is pretty irrelevant for the demo where it posts to, so any other demo endpoint like for example the star wars api will probably do.

bizzguy:
@mweststrate I agree.  It would be nice to have both GET and POST, something with a bunch of headers and a fairly complex JSON response.  And something that is pretty stable.  I'll do some research.  I also noticed that the sample app for iOS has 3 requests while Android has 2.  Probably good to normalize that as well.

bizzguy:
This is still an issue but I do not have the permissions to change the label.

PR #5065, opened 2023-08-22
fix: Android Sample App Post Request failing with 404 error #1480

FreakWolf:
## Summary
The OkHttpClient is created with connect and read timeouts to ensure that network requests do not hang indefinitely.

Proper error handling is implemented in the onFailure callback for both GET and POST requests. Error messages are logged, making it easier to debug issues.

The endpoint URLs are updated with placeholders. You should replace these with the correct URLs you want to use for your network requests.

Please make sure to update the URLs and integrate this code into your Android app as needed.

### DIFF
--- a/android/sample/src/debug/java/com/facebook/flipper/sample/ExampleActions.java
+++ b/android/sample/src/debug/java/com/facebook/flipper/sample/ExampleActions.java
@@ -22,66 +22,72 @@ import okhttp3.Response;
 
 public final class ExampleActions {
 
-  public static void sendPostRequest(OkHttpClient client) {
-    final RequestBody formBody =
-        new FormBody.Builder().add("app", "Flipper").add("remarks", "Its awesome").build();
-
-    final Request request =
-        new Request.Builder()
-            .url("https://demo9512366.mockable.io/SonarPost")
-            .post(formBody)
+    // Create a custom OkHttpClient with timeouts
+    private static final OkHttpClient client = new OkHttpClient.Builder()
+            .connectTimeout(15, TimeUnit.SECONDS)
+            .readTimeout(15, TimeUnit.SECONDS)
             .build();
 
-    client
-        .newCall(request)
-        .enqueue(
-            new Callback() {
-              @Override
-              public void onFailure(final Call call, final IOException e) {
-                e.printStackTrace();
-                Log.d("Flipper", e.getMessage());
-              }
+    public static void sendPostRequest() {
+        final RequestBody formBody =
+                new FormBody.Builder().add("app", "Flipper").add("remarks", "Its awesome").build();
 
-              @Override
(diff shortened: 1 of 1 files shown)


passy:
Thank you! For future PRs, would you mind keeping the formatting the same? Otherwise, it's very hard to actually read the diff as the entire file has changed.

FreakWolf:
Yes I'll keep in mind not to change the formatting for the future PRs and im really sorry for the inconvenience .

android/sample/src/debug/java/com/facebook/flipper/sample/ExampleActions.java
```diff
@@ -22,66 +22,72 @@
...
-    client
-        .newCall(request)
-        .enqueue(
-            new Callback() {
-              @Override
-              public void onFailure(final Call call, final IOException e) {
-                e.printStackTrace();
-                Log.d("Flipper", e.getMessage());
-              }
+    public static void sendPostRequest() {
```
passy, line 31: https://github.com/facebook/flipper/blob/5f9000aa82cfb1ce77576d22ac608356263cd393/android/sample/src/main/java/com/facebook/flipper/sample/RootComponentSpec.java#L140

This gets called here with a client, so I would expect that call to fail now. In general, having a client as a static field does not seem great to me as it will stay in memory longer than we need it.

passy (changes requested):

android/sample/src/debug/java/com/facebook/flipper/sample/ExampleActions.java
FreakWolf: made the changes as mentioned kindly review an let me know

desktop/plugins/public/shared_preferences/src/index.tsx
```diff
@@ -258,14 +258,17 @@ export function Component() {
               }
               const values = entry.preferences;
               let newValue = value;
```
passy, line 260: This looks like it's a separate change?

android/sample/src/debug/java/com/facebook/flipper/sample/ExampleActions.java
```diff
@@ -22,66 +23,95 @@
...
-              }
-            });
-  }
-
-  public static void sendNotification() {
-    final FlipperClient client = AndroidFlipperClient.getInstanceIfInitialized();
-    if (client != null) {
-      final ExampleFlipperPlugin plugin = client.getPluginByClass(ExampleFlipperPlugin.class);
-      plugin.triggerNotification();
+   public static void sendPostRequest() {
```
passy, line 26: There is still a bunch of whitespace changed. Do you think you could try to adapt that to the current formatting style?

passy (changes requested):

Issue #5124, opened 2023-09-09
Install Plugin tab screen layout broken making unable to install plugin


dprevost-LMI:
## 🐛 Bug Report
For an unknown reason, the layout of the Install Plugin tab under More > Add Plugin looks like the one below. 
As you can see, it blocks us from installing plugins, which is quite annoying.

<img width="1401" alt="Screenshot 2023-09-09 at 8 45 17 AM" src="https://github.com/facebook/flipper/assets/77302423/7b92b187-d247-4ccf-84a2-6be5752e2f3d">

## To Reproduce
- Not sure what led to the broken layout
- Go to More > Add Plugin >  Install Plugin
- The layout is broken, and we see only three letters of the plugin names at the left of the screen

## Environment
Flipper 0.216.0
Apple M2 Pro
macOS Ventura 13.5.1

dancomanlive:
Same here, very extremely annoying.

<img width="849" alt="Screenshot 2023-09-13 at 12 05 12" src="https://github.com/facebook/flipper/assets/16872821/be7cddd0-5fd8-4710-8dca-cd84c8419fea">

gkartalis:
This is still the case on Flipper version 0.220.0

shenyu1019:
same problem！

nazmeln:
+1

gianlucalippolis:
Solved with this pull request https://github.com/facebook/flipper/pull/5161

aigoncharov:
Could you try clearing your localStorage? It works for me without any problems and the values in the local storage is the only explanation I could think of

heath-clink:
After I cleared local storage and reloaded, this screen began rendering properly again. Thanks for the tip @aigoncharov

gianlucalippolis:
Hi @hpelitebook745G2 , i fixed this bug in this PR https://github.com/facebook/flipper/pull/5161 that has been merged today.

KamilDike:
0.239.0 still a thing. What the heck

PR #5161, opened 2023-09-21
Fixed plugin installation table columns width

gianlucalippolis:
## Summary

the issue was that in the plugin installation popup table, the columns had a width of 0, making it impossible to see any data and the install/uninstall button.

This PR addresses the issue #5124 

Screenshot 2023-09-21 alle 16 46 19

After the fix

image

I resolved the issue by moving the constants declaring the column widths inside the component, which previously caused the columns to have a width of 0 in the plugin installation popup table.

### DIFF
--- a/desktop/flipper-ui/src/chrome/plugin-manager/PluginInstaller.tsx
+++ b/desktop/flipper-ui/src/chrome/plugin-manager/PluginInstaller.tsx
@@ -32,6 +32,28 @@ const {Text, Link} = Typography;
 
 const TAG = 'PluginInstaller';
 
+const columnSizes = {
+  name: '25%',
+  version: '10%',
+  description: 'flex',
+  install: '15%',
+};
+
+const columns = {
+  name: {
+    value: 'Name',
+  },
+  version: {
+    value: 'Version',
+  },
+  description: {
+    value: 'Description',
+  },
+  install: {
+    value: 'Action',
+  },
+};
+
 type PropsFromState = {
   installedPlugins: Map<string, InstalledPluginDetails>;
 };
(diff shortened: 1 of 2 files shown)


passy:
Ah nice! I've seen that myself. Thanks so much for fixing this.

gianlucalippolis:
Sure @passy ! when the 'columnSizes' object containing those values was declared outside the component, at the time the return was done we could not access those values and the column width was set to zero by default. By moving the declaration inside the component, the problem did not recur.

heath-clink:
I tried this patch and it indeed fixes the issue, but not for the reason you claim. The values are actually available, which I verified by inserting `console.log(columnSizes, columns);` into the function.

The difference is: by initializing `columnSizes` and `columns` in the render function instead of globally, with each render the child component `ManagedTable` sees new objects (even though the content is semantically the same) and so it updates as if those properties had changed. I confirmed that `PluginInstaller` renders twice when the Plugins modal is opened. So the question is, why does ManagedTable seem to break that way?

As I noted in another thread, if you close the plugin modal while the first tab is selected, the next time you open it both tabs render correctly. But if you close the modal with the second tab selected, the next time you open it the second tab's table is broken.

gianlucalippolis:
@heath-clink the PluginInstaller component renders twice due to the useNPMSearch hook changing its "rows" value. I am still investigating to see if there are any alternative solutions.

gianlucalippolis:
hello everyone @passy , I apologize for abandoning the pull request but I had no more time to work on it. I think I found the problem: in the file desktop/flipper-ui-core/src/ui/components/table/TableHead.tsx at line 114 we call the onResize function passing as parameter the width value obtained from `this.ref.getBoundingClientRect().width`. 
The problem is that in this case the prop "isResizable" is passed with value to true, so in the render function, the component 'TableHeaderColumnInteractive' will be shown instead of 'TableHeadColumnContainer'. 
The difference between the two is that we do not pass the prop 'ref' to the former, so the setRef function will not be called either. A possible solution (the one I applied) would be to check at line 114 the value of the prop 'isResizable' before calling the 'onResize' function. 
I updated the pull request with this new solution.

eidan66:
Any update on this PR?
The workaround doesn't work for me :-(

laffed:
bumping.
please let's get this in, installing plugins via this ui is essentially impossible

passy:
Thanks! I'm importing this again so someone from the team can review it.

Issue #5294, opened 2023-11-16
GRPC/Connect streaming calls are missing the request body when flipper is active


ln-12:
## 🐛 Bug Report
I am currently using [connect-kotlin](https://github.com/connectrpc/connect-kotlin), a library to use [connectRPC](https://connectrpc.com/) on Android. This is a custom protocol built around [grpc](https://grpc.io/). With grpc in general, you have non streaming calls (just like normal calls to a REST api) and streaming calls. With streaming calls, a connection to the server is opened and both sides can send data into that stream until it is closed.

The problem now is that Flipper manipulates these streaming calls in a way that makes them contain a body of 0 bytes which the server can obviously not process. Here is an example of what that looks like:

![image](https://github.com/facebook/flipper/assets/36760115/a8bb694a-c91a-4103-a561-c2a5533e1948)

As you can see, the first call fails because it has a body size of 0B. However, the second call which is exactly the same, succeeds. So I would assume that it is the result of a race condition.

I think that the code responsible for this should be [here](https://github.com/facebook/flipper/blob/edc46dc2b14c66cbbd37489388ba7115e2d439ea/android/plugins/network/src/main/java/com/facebook/flipper/plugins/network/FlipperOkhttpInterceptor.java#L91-L92). When timing is bad, connect just wants to read the request body when Flipper is performing the copy operation which has the side effect that the body is empty for a very short period of time.

In the best case, this race condition is solved as when it works, Flipper is super useful to debug streaming calls.

## To Reproduce
To reproduce, I created a fork of the connect-kotlin repository and added flipper to it [here](https://github.com/ln-12/flipper-bug-reproducer). To run, clone the repo, cd into the directory and run `$ make generate`. After that you can open Android Studio and execute `examples.android`. When the app is opened, select `Connect - bidirectional streaming` from the menu. In Flipper you will see that the request body is 0B:

<img width="785" alt="image" src="https://github.com/facebook/flipper/assets/36760115/68119682-66ec-4eda-b313-50913d571743">

## Environment
Flipper 0.236.0
connect-kotlin 0.3.1

ln-12:
I added a PR which provides at least a workaround for the issue. It is not fully solved as the request bodies cannot be inspected, but at least it works again.

PR #5338, opened 2023-12-06
Update FlipperOkhttpInterceptor to skip one shot and duplex body

ln-12:
Manipulating a one shot or duplex body breaks streaming, so we skip those bodies and instead show a static placeholder.

## Summary
As already stated in [this issue](https://github.com/facebook/flipper/issues/5294), using Flipper in combination with streaming calls does not work. Flipper manipulates the request body which leads to invalid/empty requests being sent. As shown in the official OkHttp [HttpLoggingInterceptor](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt#L198-L201) and in the tool [Chucker](https://github.com/ChuckerTeam/chucker/blob/main/library/src/main/kotlin/com/chuckerteam/chucker/internal/support/RequestProcessor.kt#L54-L61), a workaround for this issue is to simply not manipulate the body in those cases.

## Changelog
Skipping one shot and duplex request body during interception

## Test Plan
I ran my [example app](https://github.com/ln-12/flipper-bug-reproducer) from the issue before and after this change. Only when skipping those bodies it worked as with the other two tools mentioned above.

### DIFF
--- a/android/plugins/network/src/main/java/com/facebook/flipper/plugins/network/FlipperOkhttpInterceptor.java
+++ b/android/plugins/network/src/main/java/com/facebook/flipper/plugins/network/FlipperOkhttpInterceptor.java
@@ -88,10 +88,18 @@ public class FlipperOkhttpInterceptor
   @Override
   public Response intercept(Interceptor.Chain chain) throws IOException {
     Request request = chain.request();
-    final Pair<Request, Buffer> requestWithClonedBody = cloneBodyAndInvalidateRequest(request);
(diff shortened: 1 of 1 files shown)


passy:
Thank you!

passy:
It looks like internally we are on a version of OkHttp that doesn't have the `isDuplex` and `isOneShot` methods yet, so I can't land this right now.

PR #5430, opened 2024-01-18
ci: Use GITHUB_OUTPUT envvar instead of set-output command

arunsathiya:
`save-state` and `set-output` commands used in GitHub Actions are deprecated and [GitHub recommends using environment files](https://github.blog/changelog/2023-07-24-github-actions-update-on-save-state-and-set-output-commands/).

This PR updates the usage of `set-output` to `$GITHUB_OUTPUT`

Instructions for envvar usage from GitHub docs:

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter


passy:
Thanks! I've been wanting to update that for a while!

arunsathiya:
@passy One minor thing that had come to my notice: had to quote the envvar to retain spaces or special characters/treat as a single argument or value. It's done in https://github.com/facebook/flipper/pull/5430/commits/3dfa3481d6c4b7927a9af3ced932f19d3983db1b. Sorry for any troubles!

passy:
@arunsathiya It looks like something is still off for the Windows build. Could you take a look?

arunsathiya:
It seems that Powershell commands need `$env` prefix, so this commit should help https://github.com/facebook/flipper/pull/5430/commits/d8a5f2b3bb198d585cff834494a5ec9a67385737, I believe. Watching this space to see how the build works out but it needs an approval first: https://github.com/facebook/flipper/actions/runs/7628655259

passy:
The windows runner still seems unhappy, unfortunately.

arunsathiya:
I remember reviewing a bit two weeks but haven't gotten a chance to look again since. 

I'll take another look soon but would like the latest checks to run once.

PR #5541, opened 2024-03-26
Moved package from AndroidManifest.xml into build.gradle namespace

emmanuelproulx:
This is necessary because Google requires this in AGP version 8.

## Summary

The code no longer works now that we're at version 8 of AGP. This is because AGP 8 requires that the package be changed to a namespace option in build.gradle.

## Changelog

The React Native module now works with Android Gradle Plugin version 8.

## Test Plan

I imported this in my own React Native project, and I was getting a build failed: "Namespace not specified. Specify a namespace in the module's build file. See https://d.android.com/r/tools/upgrade-assistant/set-namespace for information about setting the namespace."

Then I made this one change and rebuilt my project. The build worked and my app ran correctly.


passy:
Thank you!

emmanuelproulx:
The build failed but I don't understand how this could possibly be related to my change. Any idea?

emmanuelproulx:
I see now that the build was broken days ago: https://github.com/facebook/flipper/pull/5519

emmanuelproulx:
I seem to have used the wrong button.

PR #5544, opened 2024-03-29
Forward declare the DatabaseDriver protocol

jparise:
## Summary

This allows us to remove the `DatabaseDriver.h` header import, and `DatabaseDriver.h` no longer needs to be a public header.

## Changelog

`FlipperKitDatabasesPlugin/DatabaseDriver.h` is no longer a public header.

## Test Plan

Successfully built the FlipperKitDatabasesPlugin plugin code


jparise:
Or was this intentionally imported as a way to implicitly pull `DatabaseDriver.h` into the public interface?

